PR #29, opened 2020-10-30
add resource policy to manifest and helm deployer

schrodit:

/area usability
/kind api-change
/priority normal

Fixes #

**Release note**:

```improvement operator
A new manifest deployer version `v1alpha2` has been introduced that allows the usage of policies per manifest.
```


Diaphteiros:
Just a thought: Do we need a fifth policy 'ensure'?
I'm thinking about things like the VPA, for example. 
- We need it to be there.
- We want to deploy it, if it is not there.
- If we deployed it, we want to update and delete it too.
- If we did not deploy it, we also don't want to overwrite or delete it.

Basically, it behaves like 'fallback', but it also works if it is already there but it also works if there is no landscaper identity annotation or something similar.

The problem I see is that it is not trivial to check whether something exists. For CRDs, it's simple. For a deployment, for example, this 'exists' check would have to look at the container images and not at the deployment name to check whether a certain functionality is already present in the cluster.

While I think that this policy type could be nice and helpful, it is probably hard to do. What do you think?

schrodit:
> While I think that this policy type could be nice and helpful, it is probably hard to do. What do you think?
Yes I think that mode could be also helpful. But I think `ensure` is to generic. 
So I think we should use a better naming to separate it from the other modes like fallback.

> For a deployment, for example, this 'exists' check would have to look at the container images and not at the deployment name to check whether a certain functionality is already present in the cluster.
I think this could be done really easy. just check do a get for the resource, we already do the same thing for the fallback mode.

Do you have a specific scenario in mind, when you say we need to check for pods ?
How should we know if a pod that we create (we must have a specific name and namespace to manage a resource) be created by a k8s controller as all pods get a random name.
I thnínk that such scenarios should be out of scope. If someone wants to use the ensure or fallback mode on a pod managed by a deployment, they must manage the corresponding deployment.

Diaphteiros:
Yes, that's what I thought. The only case I can currently think of is the VPA.

Diaphteiros (approved):
/lgtm

schrodit:
@Diaphteiros maybe we should track that in a separate issue?

PR #30, opened 2020-11-01
add possibility to use secret or configmap to reference dataobject data

schrodit:

/area usability
/kind enhancement
/priority normal

It is now possible to specify secrets and configmaps as source for data imports.

Secretrefs and configmap refs are not allowed in installation templates which means that can only be used in installation that are created by human operators not via aggregations.
This is validated in the default validation check of a blueprint.

the documentation will follow in a different PR where the whole installation is covered. https://github.com/gardener/landscaper/pull/32

Fixes #

**Release note**:

```improvement operator
It is now also possible to specify secrets and configmaps as source for data imports.
```


pkg/apis/core/validation/installation.go
```diff
@@ -66,11 +66,22 @@ func ValidateInstallationDataImports(imports []core.DataImport, fldPath *field.P
 
 	importNames := map[string]bool{}
 	for idx, imp := range imports {
-		if imp.DataRef == "" {
-			allErrs = append(allErrs, field.Required(fldPath.Index(idx).Child("dataRef"), "dataRef must not be empty"))
+		impPath := fldPath.Index(idx)
+
+		if imp.DataRef == "" && imp.SecretRef == nil && imp.ConfigMapRef == nil {
```
Diaphteiros, line 71: Since `dataRef` is now optional, it can be `nil`, correct? Then the check should be the same as for the other two, shouldn't it?

pkg/apis/core/validation/installation.go
```diff
@@ -66,11 +66,22 @@ func ValidateInstallationDataImports(imports []core.DataImport, fldPath *field.P
...
+			allErrs = append(allErrs, field.Required(impPath, "either dataRef, secretRef or configMapRef must not be empty"))
+		}
+
```
Diaphteiros, line 74: We should also add a check that _only one_ of `dataRef`, `secretRef`, and `configMapRef` is given.

schrodit:
@Diaphteiros PTAL

Diaphteiros (approved):
/lgtm

PR #32, opened 2020-11-02
add installations docs

schrodit:

/area documentation
/kind enhancement
/priority normal

Fixes #

**Release note**:

```improvement operator
Added documentation for installations.
```

### DIFF
--- a/docs/usage/Installations.md
+++ b/docs/usage/Installations.md
@@ -1,0 +1,463 @@
+# Installations
+
+Installations are kubernetes resources that represent instances of [Blueprints](./Blueprints.md).
+Each installation contains the state of its executed blueprint and has its own context.
+
+**Index**
+- [Structure](#basic-structure)
+- [Blueprint](#blueprint)
+- [Imports](#imports)
+  - [Data Imports](#data-imports)
+  - [Target Imports](#target-imports)
+  - [Data Mappings](#import-data-mappings)
+- [Exports](#exports)
+  - [Data Export](#data-exports)
+  - [Target Export](#target-exports)
+  - [Data Mappings](#export-data-mappings)
+- [Operations](#operations)
+
+#### Basic structure:
+```yaml
+apiVersion: landscaper.gardener.cloud/v1alpha1
+kind: Installation
+metadata:
+  name: my-installation
+spec:
+
+  blueprint:
+    ref:
+#      repositoryContext:
+#        type: ociRegistry
+#        baseUrl: eu.gcr.io/myproj
+      componentName: github.com/gardener/gardener
+      version: v1.7.2
+      kind: LocalResource
+      resourceName: gardener
+#    inline:
+#      filesystem: # vfs filesystem
+#        blueprint.yaml: 
+#          apiVersion: landscaper.gardener.cloud/v1alpha1
+#          kind: Blueprint
+#          ...
+  
+  imports:
+    data:
+    - name: "" # logical internal name
+      dataRef: "" # reference a contextified data object or a global dataobject with a '#' prefix.
+#      secretRef: # reference a secret
+#        name: ""
+#        namespace: ""
+#        key: ""
+#      configMapRef: # reference a configmap
+#        name: ""
+#        namespace: ""
+#        key: ""
+    targets:
+    - name: "" # logical internal name
+      target: "" # reference a contextified target or a global taret with a '#' prefix.
+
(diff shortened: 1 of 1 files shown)


docs/usage/Installations.md
```diff
@@ -0,0 +1,459 @@
...
+      accessKeyID: (( aws-provider-type.creds.accessKeyID ))
+      accessKeySecret: (( aws-provider-type.creds.accessKeySec ))
+```
+
+
+## Exports
+
+Exports define the data that is created by the installation.
+
+Whom can distinguish between two basic types of exports:
```
Diaphteiros, line 314: Same as above - I think this 'whom' phrase sounds weird. I suggest a similar rephrasing as mentioned above.

docs/usage/Installations.md
```diff
@@ -0,0 +1,459 @@
...
+      creds: (( aws-credentials ))
+    - type: gcp
+      creds: (( gcp-credentials ))
+```
+
+## Operations
+
+The behavior of a installation is set by using operation annotations.
+These annotations are either automatically set by the landscaper itself as part of the dafault reconiclition loop.
+Or operator can set annoations to force specific behavior.
```
Diaphteiros, line 452: `either` in one sentence and `or` in the next one sounds strange to me. And there is a typo in `reconciliation`.
Proposal 1:
```
These annotations are set automatically by the landscaper as part of the default reconciliation loop.
An operator can also set annotations manually to enforce a specific behaviour.
```
(BE vs AE - behavior or behaviour?)

Proposal 2:
```
These annotations are either set automatically by the landscaper itself as part of the default reconciliation loop, or they can be set manually by an operator to enforce a specific behaviour.
```

docs/usage/Installations.md
```diff
@@ -0,0 +1,459 @@
...
+
+`landscaper.gardener.cloud/operation`:
+  - `reconcile`: start a default reconcile on the installation
+  - `force-reconcile`: skip the reconcile/pending check and directly start a new reconcilition flow. :warning: Imports still have to be satisfied.
```
Diaphteiros, line 456: Looks better this way (in my opinion):
```
  - `force-reconcile`: skip the reconcile/pending check and directly start a new reconcilition flow.
    - :warning: Imports still have to be satisfied.
```

Diaphteiros (commented):
I didn't find any content-related issues, although those are hard to detect without a lot of cross-checking, which I didn't do.

I marked multiple typo/grammatical/phrasing issues, though. In most cases, what you wrote isn't wrong or grammatically incorrect, but just sounded somewhat strange to me and I proposed an alternative phrasing of the same meaning (hopefully) that sounded better to me. There is of course room for discussion here, if you are unhappy with my proposals.

docs/usage/Installations.md
```diff
@@ -0,0 +1,459 @@
...
+          ...
+```
+
+## Imports
+
+Imports define the data that should be used by the installation to satisfy the imports of the referenced blueprint.
+
+Whom can distinguish between two basic types of imports:
+
+1. Data imports which are used to satisfy blueprint imports defined by a scheme.
```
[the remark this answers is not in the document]
schrodit: I'm also not sure, as their are used as synonyms.

docs/usage/Installations.md
schrodit: I think we should stay with one form. therefore will go with AE

schrodit:
@Diaphteiros thanks for the detailed review.
Some sentences make no sense after a lot of writing:)

I applied your comments. PTAL

docs/usage/Installations.md
Diaphteiros: Just a short question, since this is the only part that you didn't change - did you skip this one on purpose or did you miss it?

Diaphteiros:
> @Diaphteiros thanks for the detailed review.
> Some sentences make no sense after a lot of writing:)
> 
> I applied your comments. PTAL

@schrodit Yes, I know that problem.

It seems that you missed one sentence, see the only unresolved convo above. Or did you leave that out on purpose?

docs/usage/Installations.md
schrodit: Just missed that one

schrodit:
@Diaphteiros I just missed that comment.
Added it now.

Diaphteiros (approved):
/lgtm

PR #43, opened 2020-11-16
check if parent not nil

achimweigel:

/kind bug
/priority normal

For root installations there is no parent installation. This must be considered during import validation. 

**Release note**:

```improvement operator

```


schrodit (approved):
/lgtm

thanks for the fix

PR #44, opened 2020-11-16
fix checkTargetImportIsOutdated()

hendrikKahl:
/kind bug
/priority normal

Fix a bug where a `data` instead of a `target` object was accessed to check, if the import is outdated or not. 

In a scenario with an updated `target` resource, this let to a endless loop of reconciliation and the installation could never move to `succeeded` status.

Fixes #

**Release note**:

```improvement operator

```


schrodit (approved):
/lgtm

PR #47, opened 2020-11-17
Tutorial/local setup

schrodit:

/area documentation
/priority normal

Added documentation on how to setup a local landscaper with a local oci registry using harbor.

**Release note**:

```improvement operator
NONE
```


docs/tutorials/00-local-setup.md
```diff
@@ -0,0 +1,163 @@
...
+helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx
+helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+helm repo update
+
+# download the helm artifact locally
+helm pull ingress-nginx/ingress-nginx --untar --destination /tmp
+
+# upload the oci artifact to a oci registry
+export HELM_EXPERIMENTAL_OCI=1
+# helm registry login -u myuser localhost:5000 # optional if configured in the values.yaml
```
hendrikKahl, line 90:
```suggestion
# helm registry login -u admin https://my-host  # take harborAdminPassword from values.yaml, if no other user is specified
```

docs/tutorials/00-local-setup.md
```diff
@@ -0,0 +1,163 @@
...
+  enabled: false
+trivy:
+  enabled: false
+notary:
+  enabled: false
+```
+
+```
+helm repo add harbor https://helm.goharbor.io
+helm upgrade --install -n ls-system registry harbor/harbor -f values.yaml
```
MrBatschner, line 34: Do we want to install harbor into ls-system or better into a dedicated namespace?

hendrikKahl (changes requested):
some suggestions inline :) 

I guess we could add a few more hints to avoid common pitfalls. Here's a proposal:

### Working with the landscaper-cli
In order for the `landscaper-cli` to work with the registry, it needs valid credentials. The easiest way to generate these, would be via `docker login`. Here is an example for a registry accessible via port-forwarding:
```shell
docker login -u my-user localhost:5000 # use the user name and pwd as specified in the harbor chart values.yaml
```

Later, when dealing with artifacts like the component descriptor, be aware that the URLs used to push and access the artifacts differ due to the port-forwarding. Make sure the base URL points to the cluster-internal representation of the registry:

```yaml
  repositoryContexts:
  - type: ociRegistry
    baseUrl: harbor-harbor-registry.harbor:5000/comp
```
But push explicitly to `localhost` instead implicitly using the baseUrl:

```shell
landscaper-cli  componentdescriptor push localhost:5000/comp/ github.com/gardener/landscaper/ingress-nginx v0.1.0 component-descriptor.yaml
```

### Targets
For test purposes it might be feasible to deploy the landscaper as well as installations into the same (local) cluster. When creating a `target` resource, make sure the `clusters[].server` points to the cluster-internal representation of the API server (i.e `kubernetes.default.svc.cluster.local.`).

docs/tutorials/00-local-setup.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+Helm installs the registry in the `ls-system` namespace.
+The registry can be used by the landscaper by referencing artifacts via it's internal service `harbor.ls-system:80`
```
MrBatschner, line 38: At least with the helm installation above, there is no service with the name `harbor.<namespace>`.  I believe that should be `registry-harbor-registry.<namespace>:5000`.

docs/tutorials/00-local-setup.md
MrBatschner, line 90: Login should be to `https://my-host` instead of localhost.

MrBatschner (changes requested):
Some requests for changes...

docs/tutorials/00-local-setup.md
schrodit: good point, as the harbor deploys some resources we may should suggest to deploy it to the `harbor` ns.

schrodit:
@hendrikKahl @MrBatschner thanks for the comments.
I incorporated all suggestions. PTAL

MrBatschner (approved):

hendrikKahl (approved):

PR #53, opened 2020-11-24
check observed generation of di

achimweigel:
In the [reconcile method](https://github.com/gardener/landscaper/blob/c0336d6cbfcca046df3b0a6495a9766d0e04e996/pkg/landscaper/execution/reconcile.go#L43) the execution controller considers only the phase of the deploy item. It must also compare generation and observed generation of the deploy item? In the case that we observed, the deploy item spec was updated, but the deployer had not yet done anything. The deploy item phase was still successful from before. The execution controller ignored the generation difference and reported a success too early.


schrodit (approved):
/lgtm

schrodit (changes requested):
a unit test is failing.
@achimweigel can you also adjust the unit test

schrodit (approved):
/lgtm

PR #58, opened 2020-11-26
Reduce length of label values

robertgraeff:
Data objects have labels `data.landscaper.gardener.cloud/context`and `data.landscaper.gardener.cloud/source` whose values could be longer than 63 characters. This PR tries to reduce their length by
- not including the namespace
- abbreviating "Installation" (resp "Execution") to "Inst" (resp. "Exec")  

Further checks to control name lengths could be implemented later, for example in a validating webhook.

### DIFF
--- a/pkg/apis/core/v1alpha1/helper/dataobjects.go
+++ b/pkg/apis/core/v1alpha1/helper/dataobjects.go
@@ -22,6 +22,10 @@ const SourceDelimiter = "/"
 
 const NonContextifiedPrefix = "#"
 
+const InstallationPrefix = "Inst."
+
+const ExecutionPrefix = "Exec."
+
 // GenerateDataObjectName generates the unique name for a data object exported or imported by a installation.
 // It returns a non contextified data name if the name starts with a "#".
 func GenerateDataObjectName(context string, name string) string {
@@ -60,10 +64,10 @@ func ObjectFromDataObjectSource(src string) (string, lsv1alpha1.ObjectReference,
 
 // DataObjectSourceFromInstallation returns the data object source for a Installation.
 func DataObjectSourceFromInstallation(src *lsv1alpha1.Installation) string {
-	return fmt.Sprintf("Inst.%s", src.GetName())
+	return InstallationPrefix + src.GetName()
 }
 
 // DataObjectSourceFromExecution returns the data object source for a Execution.
 func DataObjectSourceFromExecution(src *lsv1alpha1.Execution) string {
-	return fmt.Sprintf("Exec.%s", src.GetName())
+	return ExecutionPrefix + src.GetName()
 }
--- a/pkg/apis/core/v1alpha1/helper/dataobjects.go
+++ b/pkg/apis/core/v1alpha1/helper/dataobjects.go
@@ -22,10 +22,10 @@ const SourceDelimiter = "/"
 
 const NonContextifiedPrefix = "#"
 
-// Prefix combined with installation name is used as label value. Do not change length.
+// InstallationPrefix is the prefix combined with installation name is used as label value. Do not change length.
 const InstallationPrefix = "Inst."
 
-// Prefix combined with execution name is used as label value. Do not change length.
+// ExecutionPrefix is the prefix combined with execution name is used as label value. Do not change length.
 const ExecutionPrefix = "Exec."
 
 // GenerateDataObjectName generates the unique name for a data object exported or imported by a installation.
--- a/pkg/apis/core/v1alpha1/helper/dataobjects.go
+++ b/pkg/apis/core/v1alpha1/helper/dataobjects.go
@@ -60,10 +60,10 @@ func ObjectFromDataObjectSource(src string) (string, lsv1alpha1.ObjectReference,
 
 // DataObjectSourceFromInstallation returns the data object source for a Installation.
 func DataObjectSourceFromInstallation(src *lsv1alpha1.Installation) string {
-	return fmt.Sprintf("Installation.%s.%s", src.GetNamespace(), src.GetName())
+	return fmt.Sprintf("Inst.%s", src.GetName())
 }
 
 // DataObjectSourceFromExecution returns the data object source for a Execution.
 func DataObjectSourceFromExecution(src *lsv1alpha1.Execution) string {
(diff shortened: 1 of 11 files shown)


schrodit (changes requested):
can you also add a restriction for the installation name in the installation validation in `pkg/apis/core/validation`?

pkg/apis/core/validation/installation.go
```diff
@@ -6,19 +6,41 @@ package validation
...
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/gardener/landscaper/pkg/apis/core/v1alpha1/helper"
+
 	"github.com/gardener/landscaper/pkg/apis/core"
 )
 
+const InstallationNameMaxLength = validation.DNS1123LabelMaxLength - len(helper.InstallationPrefix)
+
+const InstallationGenerateNameMaxLength = InstallationNameMaxLength - 5
```
schrodit, line 20: can you add a comment why the max installation name is substracted by 5.
Something like `max length of an installation - the number of random characters kubernetes uses to genreate a name`

pkg/landscaper/installations/subinstallations/subinstallations.go
```diff
@@ -151,7 +153,13 @@ func (o *Operation) createOrUpdateNewInstallation(ctx context.Context, inst *lsv
 
 	if subInst == nil {
 		subInst = &lsv1alpha1.Installation{}
-		subInst.GenerateName = fmt.Sprintf("%s-", subInstTmpl.Name)
+
+		generateName := subInstTmpl.Name
```
schrodit, line 157: I think we should inaddition also restrict the subinstallation name.

schrodit (changes requested):

pkg/apis/core/validation/installation.go
achimweigel: done

pkg/landscaper/installations/subinstallations/subinstallations.go
achimweigel: We have restricted the length of generateName of a subinstallation. There seems to be no direct setting of the name.

pkg/landscaper/installations/subinstallations/subinstallations.go
```diff
@@ -151,7 +153,13 @@ func (o *Operation) createOrUpdateNewInstallation(ctx context.Context, inst *lsv
...
+		if len(generateName) > validation.InstallationGenerateNameMaxLength-1 {
+			generateName = generateName[:validation.InstallationGenerateNameMaxLength-1]
+		}
+
+		subInst.GenerateName = fmt.Sprintf("%s-", generateName)
 		subInst.Namespace = inst.Namespace
```
achimweigel, line 163: here is the modification of the generateName

pkg/apis/core/validation/installation.go
```diff
@@ -6,19 +6,42 @@ package validation
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/gardener/landscaper/pkg/apis/core/v1alpha1/helper"
+
 	"github.com/gardener/landscaper/pkg/apis/core"
 )
 
+const InstallationNameMaxLength = validation.DNS1123LabelMaxLength - len(helper.InstallationPrefix)
```
schrodit, line 18: Please also add a comment for exported constants

pkg/apis/core/validation/installation.go
```diff
@@ -6,19 +6,42 @@ package validation
...
+
+// max length of an installation name minus the number of random characters kubernetes uses to generate a unique name
+const InstallationGenerateNameMaxLength = InstallationNameMaxLength - 5
```
schrodit, line 21:
```suggestion
// InstallationGenerateNameMaxLength is the max length of an installation name minus the number of random characters kubernetes uses to generate a unique name
const InstallationGenerateNameMaxLength = InstallationNameMaxLength - 5
```

pkg/landscaper/installations/subinstallations/subinstallations.go
schrodit: this cuts of too long names. 
I meant that we could also add the same restriction to installation templates as for installations.
So we could add 
```
if len(objMeta.GetName()) > InstallationNameMaxLength {
		allErrs = append(allErrs, field.Invalid(fldPath.Child("name"), objMeta.GetName(), validation.MaxLenError(InstallationNameMaxLength)))
	} else if len(objMeta.GetGenerateName()) > InstallationGenerateNameMaxLength {
		allErrs = append(allErrs, field.Invalid(fldPath.Child("generateName"), objMeta.GetGenerateName(), validation.MaxLenError(InstallationGenerateNameMaxLength)))
	}
```
 to https://github.com/gardener/landscaper/blob/master/pkg/apis/core/validation/blueprint.go#L262

schrodit (changes requested):

pkg/apis/core/v1alpha1/helper/dataobjects.go
```diff
@@ -22,6 +22,12 @@ const SourceDelimiter = "/"
 
 const NonContextifiedPrefix = "#"
 
+// Prefix combined with installation name is used as label value. Do not change length.
+const InstallationPrefix = "Inst."
+
+// Prefix combined with execution name is used as label value. Do not change length.
+const ExecutionPrefix = "Exec."
```
schrodit, line 29:
```suggestion
// InstallationPrefix is the prefix combined with installation name is used as label value. Do not change length.
const InstallationPrefix = "Inst."

// ExecutionPrefix is the prefix combined with execution name is used as label value. Do not change length.
const ExecutionPrefix = "Exec."
```

see the godoc style guide https://blog.golang.org/godoc

schrodit (changes requested):

achimweigel:
We added checks for the name length of subinst templates. Not sure if this was really needed because the fix before resolves problems with longer names during runtime.

schrodit (commented):
look good now.
Just one comment is missing https://github.com/gardener/landscaper/pull/58/files#r532401796

achimweigel:
commet added

schrodit (approved):
/lgtm

PR #68, opened 2020-12-07
Prevent that installation phase switches too early to Succeeded

robertgraeff:
/priority normal
/kind bug

It sporadically happens, that an installation switches to phase `Succeeded`, although an underlying deploy item is not finished. This happens sometimes after an installation or an import parameter value was changed.

- Suppose an import parameter of an installation changes.  
- The installation switches to `Progressing` and updates its execution.  
- The installation controller gets further events caused by:  
  1. the own status change. 
  2. the status change of the execution, namely when the execution controller updates the deploy item and sets the phase of the execution to `Progressing`.  
- Suppose event (i) happens before event (ii). Then the installation controller sees the execution as `Succeeded` and switches itself to `Succeeded`, which is too early.  

This PR does not only consider the phase of the execution, but also compares its generation and observed generation.


schrodit (approved):
/lgtm

PR #74, opened 2020-12-15
include component-cli with the outsourced oci client

### LINKED ISSUE #37: Factor out cnudie/component-descriptor specific implementation

/kind cleanup
/priority normal

Currently the landscaper repo includes the go specific code for interacting with the component descriptor.
This library is not landscaper specific and should be offered as separate library either in the component-spec repo or in an additional repo

In addition it should be possible to manipulate component descriptors so that blobs can automatically added in the CTF format.

schrodit:

/area dev-productivity
/kind technical-debt
/priority normal

The oci client has been outsourced to https://github.com/gardener/component-cli/tree/main/ociclient .
To avoid double maintenance and keep the oci client consistent the landscaper oci client is removed and does now rely on the central maintained client.

Basically this PR only updates the import to the new module as the rest is just a copy of the current oci client.

But with that change also the component-spec is updates which includes a breaking change:
- the component descriptor ref was previously constructed with `<repocontext base url>/<component name>:<component version>`. Now the component descriptor ref is updated to comply with the ci infrastructure that adds a component descriptor specific namespace so that the ref is now constructed with `<repocontext base url>/component-descriptor/<component name>:<component version>`
All tutorial resources have been adjusted accordingly.

Fixes #37 

Most changes are just vendor updates.
please ignore these changes

@MrBatschner and @hendrikKahl the new oci client also includes a improved oci keyring that can also handle different credentials for oci registries that are distinguished by path not only by hostname (.e.g google's container registry).
Therefore the container deployer secret sync has been adjusted

**Release note**:

```other developer
The oci client has been replaced with the centrally maintained client in https://github.com/gardener/component-cli/tree/main/ociclient.
```
```other developer
The new oci client also includes a improved oci keyring which replaces the currently custom logic in the container deployer sync secret function.
```
```breaking user
The component-descriptor library has been updated which included a breaking change. The component-descriptor oci reference path construction has been updated to now use the following format `<repocontext base url>/component-descriptor/<component name>:<component version>` instead of the old `<repocontext base url>/<component name>:<component version>`. 
```


schrodit:
cc @gardener/landscaperctl-maintainers

hack/upload-tutorial-resources.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if ! command -v landscaper-cli &> /dev/null
+then
+    echo "landscaper-cli could not be found"
+    echo "install it by running: make install-cli"
+    exit
+fi
+
+blueprints=(
+  "eu.gcr.io/gardener-project/landscaper/tutorials/blueprints/ingress-nginx:v0.2.1,./docs/tutorials/resources/ingress-nginx/blueprint"
```
hendrikKahl, line 20: should we make the version configurable?

schrodit: You mean configurable by a a argument oder through a variable?
 
Making it fully configurable is not that easily possible as other resources still need to be adjusted.
 in the component descriptor and .
e.g 
- the installation: https://github.com/gardener/landscaper/blob/master/docs/tutorials/resources/ingress-nginx/installation.yaml#L16
- the component descriptor: https://github.com/gardener/landscaper/blob/master/docs/tutorials/resources/ingress-nginx/component-descriptor.yaml#L28

hendrikKahl: I guess making it configurable via an argument to the upload script would be a good first step.

schrodit: don't know if configuration through an argument makes sense.
We would have to add at least 2 vars for the nginx and the other blueprints.

I think there is no benefit for exposing these variables as you would have to edit some resources in the repo anyway.
If It hink it makes more sense to set these versions as variable at the beginning of the script to make it more readable.

Then we would have the versions aligned in one repo and someone that runs the script would upload the approriate versions.

hendrikKahl: 👍

schrodit: @hendrikKahl PTAL

hendrikKahl: looks good, thanks!

PR #79, opened 2020-12-16
add blueprint definition to templating import values

schrodit:

/area usability
/kind enhancement
/priority normal

An additional field has been added that contains information about the currently used blueprint (e.g blueprint ref/component descriptor ref).

**Release note**:

```feature user
An additional input value for the deploy executions has been added that contains information about the blueprint
```

### DIFF
--- a/docs/usage/Blueprints.md
+++ b/docs/usage/Blueprints.md
@@ -129,7 +129,7 @@ subinstallations:
 
 A Blueprint's deploy executions may contain any number of template executors. 
 A template executor must return a list of deploy items templates.<br>
-A deploy item template exposes specific deploy item fields and will be rendered to DeployItem CRs by teh landscaper.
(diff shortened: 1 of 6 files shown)


achimweigel:
This PR contains another change in the CLI coding. Would it not be better to make such changes in the landscapercli project from now on? Of course we also could integrate this change after the PR is merged.

schrodit:
> This PR contains another change in the CLI coding. Would it not be better to make such changes in the landscapercli project from now on? Of course we also could integrate this change after the PR is merged.
Yes you're right.
I will create a PR on the landscaper-cli repo. And will also add the other changes from the component-cli PR

achimweigel:
I have already merged the changes from the component-cli PR :-)

schrodit:
then I will a PR as soon as this PR is merged

PR #83, opened 2020-12-18
[Fix]- Helm deployer is failing with nil reference which crash the whole landscaper

Nikolay-Neykov:
/area helm-deployer
/kind bug
/priority normal

If helm deployment(handled by helm-deployer) fails during deployment(for example not correct configuration of the helm chart) the ProviderStatus of the Helm object is nil. After that if we try to delete the installation containing the deployment we end up with a broken landscaper with the following error:
```Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)```
This results to broken landscaper and all finalisers are stuck. The execution, installation, deploy-items are stuck and cannot be deleted from the cluster. Even reinstall of the landscaper don't fix the error as the resources are still in the cluster.
The only way to delete the resources is to edit the resources and remove the finalisers manually.

I'm open for further discussion on the topic if needed.

**Steps to reproduce**
1. Create installation which contains helm deployment
2. Ensure invalid configuration in the helm values so that the deployment is not successfully reconciled.
3. Delete the installation
4. Result is **Status: CrashLoopBackOff** of landscaper with **invalid memory address or nil pointer dereference** error in the logs of landscaper

```other operator
- category: bugfix
Fixing successful cleanup of resources of failed helm deployment
```

### DIFF
--- a/pkg/deployer/helm/ensure.go
+++ b/pkg/deployer/helm/ensure.go
@@ -162,10 +162,15 @@ func (h *Helm) createOrUpdateExport(ctx context.Context, values map[string]inter
 }
 
 func (h *Helm) DeleteFiles(ctx context.Context) error {
+	h.log.Info("Deleting files.")
 	h.DeployItem.Status.Phase = lsv1alpha1.ExecutionPhaseDeleting
 	status := &helmv1alpha1.ProviderStatus{}
-	if _, _, err := serializer.NewCodecFactory(Helmscheme).UniversalDecoder().Decode(h.DeployItem.Status.ProviderStatus.Raw, nil, status); err != nil {
-		return err
+
+	if h.DeployItem.Status.ProviderStatus != nil {
+		h.log.V(5).Info(fmt.Sprintf("Provider status is: %s", h.DeployItem.Status.ProviderStatus))
+		if _, _, err := serializer.NewCodecFactory(Helmscheme).UniversalDecoder().Decode(h.DeployItem.Status.ProviderStatus.Raw, nil, status); err != nil {
+			return err
+		}
 	}
 
 	if len(status.ManagedResources) == 0 {


gardener-robot-ci-2:
Thank you @Nikolay-Neykov for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/deployer/helm/ensure.go
```diff
@@ -162,10 +162,25 @@ func (h *Helm) createOrUpdateExport(ctx context.Context, values map[string]inter
...
+	decoder := factory.UniversalDecoder()
+	if &decoder == nil {
+		return errors.New("Cannot get decoder from factory")
+	}
+
+	if h.DeployItem.Status.ProviderStatus != nil {
+		h.log.Info(fmt.Sprintf("Provider status is: %s", h.DeployItem.Status.ProviderStatus))
+		if _, _, err := decoder.Decode(h.DeployItem.Status.ProviderStatus.Raw, nil, status); err != nil {
+			return err
+		}
```
schrodit, line 183: the factory and the decoder can never be nil therefore the check is not needed.
```suggestion
	if h.DeployItem.Status.ProviderStatus != nil {
	        serializer.NewCodecFactory(Helmscheme).UniversalDecoder()
		h.log.Info(fmt.Sprintf("Provider status is: %s", h.DeployItem.Status.ProviderStatus))
		if _, _, err := decoder.Decode(h.DeployItem.Status.ProviderStatus.Raw, nil, status); err != nil {
			return err
		}
```

pkg/deployer/helm/ensure.go
```diff
@@ -162,10 +162,25 @@ func (h *Helm) createOrUpdateExport(ctx context.Context, values map[string]inter
...
+		return fmt.Errorf("Cannot create factory for helm scheme %s", Helmscheme)
+	}
+
+	decoder := factory.UniversalDecoder()
+	if &decoder == nil {
+		return errors.New("Cannot get decoder from factory")
+	}
+
+	if h.DeployItem.Status.ProviderStatus != nil {
+		h.log.Info(fmt.Sprintf("Provider status is: %s", h.DeployItem.Status.ProviderStatus))
```
schrodit, line 180: Can you increase the logging verbosity. Otherwise the logs would be filled up with the provider status.
e.g.
```suggestion
		h.log.V(5).Info(fmt.Sprintf("Provider status is: %s", h.DeployItem.Status.ProviderStatus))
```

schrodit (changes requested):
Thanks for the nice PR.

I just have some minor nits and can you sign the cla?

pkg/deployer/helm/ensure.go
Nikolay-Neykov: Applied.

pkg/deployer/helm/ensure.go
Nikolay-Neykov: Applied.

Nikolay-Neykov:
Tim, thanks for the prompt response. I've applied you feedback and for the CLA I've signed it and clicked "recheck" on the CLAassistant but it seems that it is still not detected. Do I need to do something else to trigger recheck?

schrodit:
@Nikolay-Neykov you have to update the author of the commits otherwise the bot will not be able to match your signed cla.
You can also sign the CLA with your other account

Nikolay-Neykov:
Done. I've fixed the commit Author.

Nikolay-Neykov:
@schrodit I see that the validations of the PR are broken, but I cannot see the reason because of 401 Unauthorised. How can I get permissions for the concourse?

schrodit:
/ok-to-test

@Nikolay-Neykov it does not look broken to me. It looks rather not tested.

schrodit (approved):

PR #85, opened 2021-01-04
Add terraform deployer

### LINKED ISSUE #40: Terraform Deployer

/kind enhancement
/priority normal

Currently , developers are able to deploy helm charts, kubernetes manifests or run custom code to install their software.
Some components may also need infrastructure resources, which are mainly deployed using Terraform.

Therefore, an additional deployer that is able to handle terraform is a good shared deployer.

I would be good if we could leverage the already existing [gardener terraformer](https://github.com/gardener/terraformer) implementation.

The output should be the terraform output as json.
The inputs may need to be adjusted.

```yaml
apiVersion: landscaper.gardener.cloud/v1alpha1
kind: DeployItem
metadata:
  name: my-bucket
spec:
  type: landscaper.gardener.cloud/terraform
  
  target: # check what is the correct target type for terraform resources

  config:
    apiVersion: terraform.deployer.landscaper.gardener.cloud/v1alpha1
    kind: ProviderConfiguration
    
    main.tf: {} 
    variables.tf: {} 
    terraform.tfvar: {}
```

spiarh:

/kind enhancement
/priority normal

This commit adds a new deployer of type terraform which
allows one to manage infrastructure resources using terraform.

Managing infra is done by running a Terraformer Pod v2
(gardener/terraformer). The name of the pod is unique which
guarantees that two terraform commands can't be run at the same time.

During a reconciliation, a new pod will be started, later on check
during the reconcile loop and requeued if the pod is still running.
This allows managing different infrastructure in parallel considering
it can take several minutes per execution.

To bump controller-runtime version, migrate AddFinalizer and
RemoveFinalizer from metav1.Object to controllerutil.Object

Signed-off-by: lcavajani <lcavajani@suse.com>

Fixes #40 

This is a follow-up on https://github.com/gardener/landscaper/pull/62

I've created a new PR because it was cleaner than rebasing the old one which changed a lot since.

I have addressed the requested changes from the first PR:
* It does not relies on the Terraformer extension from gardener/gardener
* It now relies on the image based from gardener/terraformer to create the deployer pod, it does not require it as a golang package.
* The Terraformer configuration is not exposed anymore in the DeployItem
* I have removed the extra kubeconfig and external secret references

I've left the `TargetSelector` as part of the Item but they are not used until a correct Target for this `DeployItem` can be used.

**Release note**:

```improvement operator
Add a new Terraform Deployer to manage infrastructure resources using terraform.
```


gardener-robot-ci-2:
Thank you @lcavajani for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/apis/deployer/terraform/types_config.go
```diff
@@ -0,0 +1,37 @@
...
+	// Namespace is the namespace where the Terraformer will run and store the resources.
+	// The namespace will be defaulted to the default namespace.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+	// Image is the name of the terraformer container image.
+	// The image will be defaulted by the terraform deployer to the configured default.
+	// +optional
+	Image string `json:"image,omitempty"`
+	// LogLevel is the log level of the terraformer.
+	// Default to info.
```
schrodit, line 34: its defaulted to `error` in the defaulter.
But I think we should set it to info as defined in your comment.

schrodit: maybe this is also something we could make configurable in the provider configuration

pkg/apis/deployer/terraform/v1alpha1/defaults.go
```diff
@@ -0,0 +1,27 @@
...
+	return RegisterDefaults(scheme)
+}
+
+// SetDefaults_Configuration sets the defaults for the terraform deployer configuration.
+func SetDefaults_Configuration(obj *Configuration) {
+	if len(obj.Terraformer.Namespace) == 0 {
+		obj.Terraformer.Namespace = metav1.NamespaceDefault
+	}
+	if len(obj.Terraformer.Image) == 0 {
+		obj.Terraformer.Image = "eu.gcr.io/gardener-project/gardener/terraformer:v2.0.0"
```
schrodit, line 22: can you add a todo to add a component reference to the terraformer component in the future.

pkg/deployer/terraform/actuator.go
```diff
@@ -0,0 +1,179 @@
...
+	// OperationDelete is the value of the Operation that defines a delete operation.
+	OperationDelete OperationType = "DELETE"
+)
+
+func NewActuator(log logr.Logger, config *terraformv1alpha1.Configuration) (reconcile.Reconciler, error) {
+	return &actuator{
+		log:    log,
+		config: config,
+	}, nil
+}
```
schrodit, line 43: I guess these are not needed in the terraformer deployer?

pkg/deployer/terraform/ensure.go
```diff
@@ -0,0 +1,215 @@
...
+
+const (
+	// RequeueAfter is the time after which to requeue the item.
+	ItemRequeue = 15 * time.Second
+
+	// DeadLinePodNotReady is the time to wait before deleting a not ready pod.
+	DeadLinePodNotReady = 10 * time.Minute
+)
+
+func (t *Terraform) Reconcile(ctx context.Context, op OperationType) (reconcile.Result, error) {
```
schrodit, line 38: this method looks a lot like the code in the container deploxer.
Maybe we can share some code for pod reconcilition etc.

pkg/apis/deployer/terraform/v1alpha1/types_config.go
```diff
@@ -0,0 +1,45 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ManagedInstanceLabel describes label that is added to every terraform deployer managed resource
+// to define its corresponding instance.
+const ManagedInstanceLabel = "terraform.deployer.landscaper.gardener.cloud/instance"
+
+// ManagedDeployItemLabel describes label that is added to every terraform deployer managed resource
+// to define its source deploy item.
+const ManagedDeployItemLabel = "terraform.deployer.landscaper.gardener.cloud/deployitem"
```
schrodit, line 18: there is also a constants file in `pkg/deployers/terraform`. We should have only one place where constants are defined

schrodit (commented):
Thanks for the terraform deployer.
Looks really good.
Will have deeper look tomorrow and will test it out.

/ok-to-test

pkg/apis/deployer/terraform/types_config.go
lcavajani: > its defaulted to `error` in the defaulter.
> But I think we should set it to info as defined in your comment.

yes it was supposed to `info` in the defaulter as well.

lcavajani: > maybe this is also something we could make configurable in the provider configuration

This is specific to the Terraformer itself so this would be visible in the controller logs and when a Terraform command fails.

pkg/deployer/terraform/actuator.go
lcavajani: `OperationType` is used in `reconcile` to change the logic of the Terraform reconciliation. By default it will run `apply` or `destroy` with `OperationDelete`. But there is a safe guard to avoid deleting a stack that did not succeed before.

pkg/apis/deployer/terraform/v1alpha1/types_config.go
lcavajani: The constants there are specifics to the controller logic, should I move that to a `pkg/apis/deployer/terraform/constants.go` ?

lcavajani:
@schrodit I could not check in concourse why the previous build failed. I have made a few changes and rebased in the meantime.

schrodit:
I will retrigger the pipeline
/ok-to-test

pkg/apis/deployer/terraform/v1alpha1/types_config.go
schrodit: yes I that makes sense.
Then we have 1 file were all constants for the terraform deployer is defined

Diaphteiros (commented):
Just one small thing I noticed, lgtm otherwise.

Diaphteiros:
I just noticed a problem that we should talk about:
We use the [Gardener terraformer](https://github.com/gardener/terraformer). That comes with specific terraform providers pre-installed, there are multiple images containing different providers. The most conclusive one contains the providers listed [here](https://github.com/gardener/terraformer/blob/master/build/all/terraform-bundle.hcl#L10-L18).

The terraformer then calls the `terraform init` command with the `-plugin-dir` argument pointing to the pre-installed providers. However, this argument also disables dynamic loading of missing providers, thus limiting the terraformer to the pre-installed ones. The argument is hard-coded, see here: https://github.com/gardener/terraformer/blob/master/pkg/terraformer/terraformer.go#L186

While this is no problem for the Gardener itself, were the terraformer is used within a defined scope, I don't think we want to limit the users of our terraform deployer to these providers.

My proposal would be to open a PR on the terraformer repo and change two things:
1. Make the `-plugin-dir` argument to `terraform init` configurable.
2. Add another docker image option that comes without pre-installed images - since we need dynamic loading anyway, there is no need to use a bigger image than necessary.

@schrodit WDYT? Maybe we should move this discussion into an issue?
(I came across this problem while trying to use the `kubernetes` plugin as you proposed for e2e tests - this provider is missing.)

schrodit:
> I just noticed a problem that we should talk about:
> We use the [Gardener terraformer](https://github.com/gardener/terraformer). That comes with specific terraform providers pre-installed, there are multiple images containing different providers. The most conclusive one contains the providers listed [here](https://github.com/gardener/terraformer/blob/master/build/all/terraform-bundle.hcl#L10-L18).
> 
> The terraformer then calls the `terraform init` command with the `-plugin-dir` argument pointing to the pre-installed providers. However, this argument also disables dynamic loading of missing providers, thus limiting the terraformer to the pre-installed ones. The argument is hard-coded, see here: https://github.com/gardener/terraformer/blob/master/pkg/terraformer/terraformer.go#L186
> 
> While this is no problem for the Gardener itself, were the terraformer is used within a defined scope, I don't think we want to limit the users of our terraform deployer to these providers.
> 
> My proposal would be to open a PR on the terraformer repo and change two things:
> 
> 1. Make the `-plugin-dir` argument to `terraform init` configurable.
> 2. Add another docker image option that comes without pre-installed images - since we need dynamic loading anyway, there is no need to use a bigger image than necessary.
> 
> @schrodit WDYT? Maybe we should move this discussion into an issue?
> (I came across this problem while trying to use the `kubernetes` plugin as you proposed for e2e tests - this provider is missing.)

Oh yes really good point. And we definitly need the init functionality.

So yes the only way I see is to open a PR in the terraform repository and maybe buildt another image where the dynamic loading can be used

pkg/deployer/terraform/terraformer/config.go
```diff
@@ -0,0 +1,150 @@
...
+				return false, nil
+			}
+			t.log.Error(err, "unable to get variables secret", "secret", variablesKey.String())
+			return false, err
+		}
+
+		configurationKey := kutils.ObjectKey(t.ConfigurationConfigMapName, t.Namespace)
+		if err = t.kubeClient.Get(pollCtx, configurationKey, &corev1.ConfigMap{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				t.log.Error(err, "unable to get configuration configmap", "configmap", configurationKey.String())
```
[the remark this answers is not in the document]
lcavajani: it was not intended, will fix it

schrodit:
will close that pr in favor of https://github.com/gardener/landscaper/pull/147 which build upon this one and adds enhancements described by @Diaphteiros .
@lcavajani hope that is fine for you? (I will keep you as co-author for the commits:))

PR #99, opened 2021-01-28
add integration test framework and first integration test

### LINKED ISSUE #34: Integration tests

/area quality
/kind test
/priority normal

To be sure that a PR is not breaking anything we need to have integration tests that cover the complete workflow with
- build blueprint + component descriptor (maybe in our own registry)
- deploy simple applications with oci artifacts (preferably the tutorial resources are used)

In addition, we should also create integration tests for deployers.

Note:It would be the best option to use kind clusters to speed up the process and have it local reproducible.
Alternativly we could leverage the testmachinery and shoot hosts

- [ ] Integration test framework to easily add tests (https://github.com/gardener/landscaper/pull/99)
  - [ ] Automatically create kind clusters
- [ ] Use tutorial resources with hardcoded blueprints and component descriptors
- [ ] Include blueprint and component descriptor upload
- [ ] Deployer integration tests
- [ ] Test component descriptor/blueprint and other artifact pull from private registries

schrodit:

/area testing
/kind test
/priority critical

Adds a integration test framework that runs tests against a running landscaper intallation.
In addition the first test has been added that tests the first simple tutorial.

Currently the tests can be executed by targeting a cluster with a running landscaper (set env var) and then run `make integration-test`.
The `make test`  command has been adjusted so that it only runs unit tests.

Currently the test can only be executed manually.
But I'm currently trying to find the best way to automatically run the test in the pipline

Partially fixes https://github.com/gardener/landscaper/issues/34

I only added one test to not make the PR that big.
Therefore the other tests will be added in a separate PR when this one is merged.

**Release note**:

```other developer
A integration test framework has been added that make it possible to run tests against a cluster with a running landscaper.
```
```other developer
A integration test has been added that tests the simple nginx-ingress tutorial.
```


test/utils/envtest/environment.go
```diff
@@ -84,11 +82,16 @@ func (e *Environment) Stop() error {
...
 func (e *Environment) InitState(ctx context.Context) (*State, error) {
+	return InitStateWithNamespace(ctx, e.Client)
+}
+
+// InitState creates a new isolated environment with its own namespace.
+func InitStateWithNamespace(ctx context.Context, c client.Client) (*State, error) {
 	state := NewState()
 	// create a new testing namespace
 	ns := &corev1.Namespace{}
 	ns.GenerateName = "unit-tests-"
```
hendrikKahl, line 93: Not sure, if it is used in a different context, shouldn't this be `integration-tests-` ?

schrodit: its also used in unit tests. So will just use `tests-` as prefix

test/utils/envtest/state.go
```diff
@@ -75,15 +75,130 @@ func (s *State) AddResources(objects ...client.Object) error {
...
+		if err := cleanupForObject(ctx, c, obj); err != nil {
+			return err
+		}
+	}
+	for _, obj := range s.ConfigMaps {
+		if err := cleanupForObject(ctx, c, obj); err != nil {
+			return err
+		}
+	}
+
```
hendrikKahl, line 169: for the sake of completeness - would it make sense to remove `dataobjects` and `targets` as well?

```suggestion
	for _, obj := range s.Targets {
		if err := cleanupForObject(ctx, c, obj); err != nil {
			return err
		}
	}
	for _, obj := range s.DataObjects {
		if err := cleanupForObject(ctx, c, obj); err != nil {
			return err
		}
	}
```

test/utils/envtest/state.go
```diff
@@ -75,15 +75,130 @@ func (s *State) AddResources(objects ...client.Object) error {
...
+	for _, obj := range s.Generic {
+		if err := cleanupForObject(ctx, c, obj); err != nil {
+			return err
+		}
+	}
+
```
hendrikKahl, line 175: A failed container-deployer pod might block the namespace deletion. To be on the safe side, maybe we could purge finalizers from pods in the namespace too? 

```suggestion
	pods := &corev1.PodList{}
	c.List(ctx, pods, &client.ListOptions{Namespace: s.Namespace})
	for _, obj := range pods.Items {
		if err := removeFinalizer(ctx, c, &obj); err != nil {
			return err
		}
	}
```

hendrikKahl (commented):
looks good overall :) I added some suggestions for the cleanup procedure.

test/utils/envtest/state.go
schrodit: ah yes makes totally sense

PR #100, opened 2021-01-28
Update controller-runtime dependency

Diaphteiros:

/area landscaper
/kind enhancement
/priority blocker

Upgrades the controller-runtime dependency to `v0.8.1`.
Also updates the k8s dependency in apis/go.mod

**Release note**:

```other dependency
Update controller-runtime dependency to `v0.8.1`
```


Diaphteiros:
The problem is probably that with `github.com/gardener/landscaper/apis v0.0.0` in the main go.mod file, the wrong version of the apis is vendored. Not sure how to resolve this, since I would need to vendor the version that I'm currently working on ...

pkg/deployer/helm/ensure.go
```diff
@@ -201,16 +201,17 @@ func (h *Helm) DeleteFiles(ctx context.Context) error {
...
-	controllerutil.RemoveFinalizer(&h.DeployItem.ObjectMeta, lsv1alpha1.LandscaperFinalizer)
+	controllerutil.RemoveFinalizer(h.DeployItem, lsv1alpha1.LandscaperFinalizer)
 	return h.kubeClient.Update(ctx, h.DeployItem)
 }
 
 // ApplyObject applies a managed resource to the target cluster.
 func (h *Helm) ApplyObject(ctx context.Context, kubeClient client.Client, obj *unstructured.Unstructured) error {
 	currOp := "ApplyObjects"
-	currObj := obj.NewEmptyInstance()
+	currObj := unstructured.Unstructured{}
```
schrodit, line 211: has `NewEmptyInstance` been removed?

schrodit (changes requested):
looks good when the test is fixed.
Can you directly do a refactoring for that todo now that the new clinet.Object is available: https://github.com/gardener/landscaper/blob/master/test/utils/envtest/types.go#L20

pkg/deployer/helm/ensure.go
Diaphteiros: No, but it returns an object of type `runtime.Unstructured` which does not implement the `client.Object` interface that is now required by `kubeClient.Get`. I'm somewhat confused why `NewEmptyInstance` called on type `unstructured.Unstructured` - which implements the interface - does return a different type. This was the first workaround that I found - it basically does exactly the same thing as `NewEmptyInstance`, but with the `unstructured.Unstructured` type.

schrodit: ah makes sense.
Can you add that as a comment maybe

Diaphteiros: done

Diaphteiros:
> looks good when the test is fixed.
> Can you directly do a refactoring for that todo now that the new clinet.Object is available: https://github.com/gardener/landscaper/blob/master/test/utils/envtest/types.go#L20

done

schrodit (approved):
looks good now 
/lgtm

PR #101, opened 2021-01-28
automatically deploy required CRDs into the cluster

### LINKED ISSUE #98: Automate CRD deployment

/area robustness
/kind enhancement
/priority normal

Currently the helm chart deploys the custom resource definitions of the landscaper.
This has some disadvantages as helm charts cannot handle specific use cases like (the crd is already installed by another landscaper in the cluster). 

Therefore, the crd's should be automatically installed by the landscaper itself as we can have more control over how the crds are provided.

MrBatschner:

/area landscaper
/kind enhancement
/priority normal

Landscaper - in particular the landscaper-controller - works on its own custom resources of which the CRDs have to be deployed to the cluster prior to usage. With this PR, the landscaper-controller will deploy/update its CRDs automatically to the cluster on startup. This behaviour is configurable through the config file.

Fixes #98 

The changes in this PR make use of the `embed` package of Go 1.16 to embed the CRD yaml files into the landscaper-controller binary. This PR can only be merged once Go 1.16 is released and the CI infrastructure has been upgraded to Go 1.16.

**Release note**:

```feature operator
The landscaper-controller will now automatically deploy the CRDs for its resources upon startup.
```


pkg/landscaper/crdmanager/crdmanager.go
```diff
@@ -0,0 +1,186 @@
...
+type CRDManager struct {
+	cfg          config.CrdManagementConfiguration
+	crdClientset *apiextensions.Clientset
+	log          logr.Logger
+	crdRawDataFS *embed.FS
+}
+
+// NewCrdManager returns a new instance of a CRDManager
+func NewCrdManager(log logr.Logger, mgr manager.Manager, lsConfig *config.LandscaperConfiguration) (*CRDManager, error) {
+	crdClientset, err := apiextensions.NewForConfig(mgr.GetConfig())
```
schrodit, line 46: I would work here with the `client.Client` package from the controller runtime to be consistent with the rest of the code.

pkg/landscaper/crdmanager/crdmanager.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		log.V(3).Error(err, "failed to setup client to register CRD's")
+		return nil, err
+	}
+
+	cfg := config.CrdManagementConfiguration{
```
schrodit, line 52: Couldn't we just set the config from ls config like `cfg := lsConfig.CrdManagement` ?

apis/config/v1alpha1/types_landscaper_config.go
```diff
@@ -67,3 +69,13 @@ type OCICacheConfiguration struct {
 	// +optional
 	Path string `json:"path"`
 }
+
+// CrdManagementConfiguration contains the configuration of the CRD management
+type CrdManagementConfiguration struct {
+	// DeployCustomResourceDefinitions specifies if CRDs should be deployed
+	DeployCustomResourceDefinitions bool `json:"deployCrd"`
```
schrodit, line 76: Should we maybe default it to `true`

pkg/landscaper/crdmanager/crdmanager.go
```diff
@@ -0,0 +1,186 @@
...
+	log          logr.Logger
+	crdRawDataFS *embed.FS
+}
+
+// NewCrdManager returns a new instance of a CRDManager
+func NewCrdManager(log logr.Logger, mgr manager.Manager, lsConfig *config.LandscaperConfiguration) (*CRDManager, error) {
+	crdClientset, err := apiextensions.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		log.V(3).Error(err, "failed to setup client to register CRD's")
+		return nil, err
```
schrodit, line 49: the error is already logged in the calling function. 
```suggestion
		log.V(3).Error(err, "failed to setup client to register CRD's")
		return nil, fmt.Errorf("failed to setup client to register CRD's: %w", err)
```

pkg/landscaper/crdmanager/crdmanager.go
```diff
@@ -0,0 +1,186 @@
...
+			crdmgr.log.V(3).Error(err, "failed to read CRD file")
+			return nil, err
+		}
+
+		decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), readerBufferSize)
+		crd := &v1.CustomResourceDefinition{}
+		err = decoder.Decode(crd)
+		if err != nil {
+			crdmgr.log.V(3).Error(err, "failed to decode CRD from file")
+			return nil, err
```
schrodit, line 179: the err should be logged by the caller and therefore it would be logged 2 times.
```suggestion
			return nil, fmt.Errorf("failed to decode CRD from file: %w", err)
```

schrodit (commented):
just some first observations.
I really like the approach with the separate crd manager to keep the logic separated:)

pkg/landscaper/crdmanager/crdmanager.go
MrBatschner: Fixed.

pkg/landscaper/crdmanager/crdmanager.go
MrBatschner: Fixed.

pkg/landscaper/crdmanager/crdmanager.go
MrBatschner: Fixed.

pkg/landscaper/crdmanager/crdmanager.go
hendrikKahl: unfortunately, `client.Client` does not include direct access to scheme `apiextensions`. The `crdClienset` uses the client-go package and wraps it conveniently. Hence, we'd prefer to stick to it :)

apis/config/v1alpha1/types_landscaper_config.go
MrBatschner: The configuration of the CRD management was meant to be mandatory but yes, it makes more sense to keep it optional with `deployCrd: true` as default. Fixed!

pkg/landscaper/crdmanager/crdmanager.go
schrodit: > unfortunately, client.Client does not include direct access to scheme apiextensions
it also only includes the "default" k8s types because they are defaulted. You can set the used schems in the clients option's:
```
import (
  apiextinstall "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/install"
 )
apiextensionsScheme := runtime.NewScheme()
apiextinstall.Install(apiextensionsScheme)
kubeClient, err := client.New(mgr.GetConfig(), client.Options{
  Scheme: apiextensionsScheme,
})
```

hendrikKahl: thanks for the hint! Switch to client.Client has been pushed.

schrodit:
hold until go 1.16 is released
/hold

MrBatschner:
Rebased to latest changes on master. Now with go 1.16, we can finally merge this PR.
/invite schrodit
@schrodit Can you have another quick look?

schrodit (approved):
looks good
/lgtm

MrBatschner:
one more time... ♫
/test

PR #117, opened 2021-02-12
add test for simple import and aggregation tutorial

schrodit:

/area testing
/kind test
/priority normal

Fixes #

**Release note**:

```other developer
A integration test for the simple-import and aggregation tutorial has been added.
```
```bugfix user
A bug has been fixed that caused the import validation to stuck when a non-dataref import has been defined on a dependent installation.
```


MrBatschner (changes requested):
Looks good to me so far. See comments below. Still need to run the test locally, so far it fails because of unobtainable DataObjects...

MrBatschner (approved):

PR #119, opened 2021-02-15
Fix error and state handling in manifest deployer

schrodit:

/area quality
/kind bug
/priority normal

A bug in the manifest deployer has been fixed that causes it to not properly write its state in case of errors during the apply.

Alos the code has been refactored to now deploy all manifests in parallel.
Most logic is now factored out to a `ManifestApplier` the idea to later use that applier for the manifest and helm deployer.
Added e2e tests for the manifest deployer

Fixes #

**Release note**:

```bugfix operator
[Manifest Deployer] A bug in the manifest deployer has been fixed that causes it to not properly write its state in case of errors during the apply.
```


pkg/deployer/manifest/test/e2e_test.go
```diff
@@ -0,0 +1,240 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		By("create deploy item")
+		di := ReadAndCreateOrUpdateDeployItem(ctx, testenv, state, "ingress-test-di", "./testdata/01-di.yaml")
+
+		// First reconcile will add a finalizer
+		testutil.ShouldReconcile(ctx, actuator, testutil.Request(di.GetName(), di.GetNamespace()))
+		testutil.ShouldReconcile(ctx, actuator, testutil.Request(di.GetName(), di.GetNamespace()))
+
+		Expect(testenv.Client.Get(ctx, testutil.Request(di.GetName(), di.GetNamespace()).NamespacedName, di)).To(Succeed())
```
Diaphteiros, line 166: The first part of each test function - nearly half of it, actually - is identical. Does it make sense to move this into some kind of `BeforeEach`? If not, can this be moved into a `prepareTest` (or something like this) helper function then? It's quite a lot of duplicate code here ...
WDYT?

Diaphteiros (changes requested):
Looks good, just one minor question regarding code duplication in the e2e tests.

pkg/deployer/manifest/test/e2e_test.go
schrodit: most of the code is already in helper functions.
Are aouy talking about the should reconcile functions?

But for the injections methods, I would rather go for https://github.com/kubernetes-sigs/controller-runtime/pull/1322
so that we do not even have these inject functions anymore.

Diaphteiros: I'm talking about this part: https://github.com/schrodit/landscaper/blob/fix/manifest-deployer/pkg/deployer/manifest/test/e2e_test.go#L45-L64

Unless I'm missing something, it's completely identical to https://github.com/schrodit/landscaper/blob/fix/manifest-deployer/pkg/deployer/manifest/test/e2e_test.go#L94-L114 and https://github.com/schrodit/landscaper/blob/fix/manifest-deployer/pkg/deployer/manifest/test/e2e_test.go#L146-L166.
Ok, the first marked area has one line less, so they are not completely identical, but still pretty much. Is there a way to reduce the code duplication here or is it better to leave it this way?

schrodit: ok yes so as said the 2 injections methods (e.g. `inject.SchemeInto`) will soon be completely removed.

So when we talk about code duplication it is
```
                 ctx := context.Background()
		defer ctx.Done()
                 actuator, err := manifestactuator.NewActuator(logtesting.NullLogger{}, &manifestv1alpha1.Configuration{})
		Expect(err).ToNot(HaveOccurred())
             
		di := ReadAndCreateOrUpdateDeployItem(ctx, testenv, state, "ingress-test-di", "./testdata/01-di.yaml")

		// First reconcile will add a finalizer
		testutil.ShouldReconcile(ctx, actuator, testutil.Request(di.GetName(), di.GetNamespace()))
		testutil.ShouldReconcile(ctx, actuator, testutil.Request(di.GetName(), di.GetNamespace()))

		Expect(testenv.Client.Get(ctx, testutil.Request(di.GetName(), di.GetNamespace()).NamespacedName, di)).To(Succeed())
```
with that code snippet I do not think that a helper method would increase the readability.
There are then 4 function calls and I would like to keep them so that someone who reads that code knows exactly what is happening:
1. create a deployitem defined as yaml from "./testdata/01-di.yaml"
2. Reconcile 2 times
3. After 2 Times of reconcile the deployitem should be succeeded

If a put that in another function it would be something like
```
ctx := context.Background()
defer ctx.Done()
actuator, err := manifestactuator.NewActuator(logtesting.NullLogger{}, &manifestv1alpha1.Configuration{})
Expect(err).ToNot(HaveOccurred())
di := ReadAndCreateDeployItemAndReconcileAndWaitForPhase(ctx, actuator, testenv, state, "ingress-test-di", "./testdata/01-di.yaml")
```

I think the first example is more readable although we duplicate some code

Diaphteiros (approved):
/lgtm

PR #143, opened 2021-03-11
Deploy Item Pickup Timeout Detection

### LINKED ISSUE #115: DeployItem Timeout

/area operations
/kind enhancement
/priority normal

Currently the landscaper waits infinite for deploy items to become ready or even to be picked up by deployers.
There should a timeout when the landscaper should assume that a deploy item will never become ready.

We should distinguish between the picked up timeout and the running timeout (which could be much higher)
We should also think of make the timeout configurable in the deploy item.

Diaphteiros:

/area operations
/kind enhancement
/priority 3

This PR adds a controller to the landscaper that watches DeployItems. By adding a timestamp annotation to new/modified DeployItems, it is able to recognize when DeployItems are not handled by any deployer ("pickup timeouts").
The existing deployers have been adapted accordingly - they will remove the timeout annotation whenever they see it to signal that this DeployItem is taken care of.

Part of #115 

**Release note**:

```feature operator
The landscaper now detects if a DeployItem is not reconciled by any deployer within a specified timeframe and puts these DeployItems in `Failed` state. The timeout duration defaults to five minutes and can be set in the landscaper configuration.
```
```breaking developer
Deployers now have to remove timeout annotations (`landscaper.gardener.cloud/reconcile-time`) and update `status.observedGeneration` at the beginning of the DeployItem reconcile loop. See the new [documentation on expected deployer behaviour](docs/technical/deployer_contract.md) for more information.
```


Diaphteiros:
Current state:
I did some simple tests with mock, container, helm, and manifest deployitems and everything behaved as expected.
We should run some more advanced tests to verify and also add integration tests for this.
Also I have seen this error sometimes when deleting a helm deployitem: 
```{"level":"error","ts":"2021-03-11T12:21:26.019+0100","logger":"controller-runtime.manager.controller.deployitem","msg":"Reconciler error","reconciler group":"landscaper.gardener.cloud","reconciler kind":"DeployItem","name":"nginx-di","namespace":"default","error":"Operation cannot be fulfilled on deployitems.landscaper.gardener.cloud \"nginx-di\": StorageError: invalid object, Code: 4, Key: /registry/landscaper.gardener.cloud/deployitems/default/nginx-di, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 2e84bde8-3546-4b41-9efb-cfe92ef645c3, UID in object meta: "}```
I'll investigate this.

File: apis/core/types_deployitem.go [hunk 1]
```diff
@@ -54,6 +54,9 @@ type DeployItemSpec struct {
 	// Note that the type information is used to determine the secret key and the type of the secret.
 	// +optional
 	RegistryPullSecrets []ObjectReference `json:"registryPullSecrets,omitempty"`
+	// PickupTimeout is the time in seconds after which the DeployItem goes into a failure state if it has not been picked up by any deployer.
+	// +optional
+	PickupTimeout *int64 `json:"pickupTimeout,omitempty"`
```
schrodit, line 59: not sure if this timeout should be configured on the deploy item.
I think the pickup timeout should be configured in the landsacper config.

When the watlk about the overall deploy item timeout, this should be configurable per deploy item

apis/core/v1alpha1/constants.go
```diff
@@ -19,6 +19,9 @@ const (
 	// OperationAnnotation is the annotation that specifies a operation for a component
 	OperationAnnotation = "landscaper.gardener.cloud/operation"
 
+	// TimeoutAnnotation is used to recognize timeouts in deployitems
+	TimeoutAnnotation = "landscaper.gardener.cloud/timeout"
```
schrodit, line 23: maybe we not specify the timeout data here but rather the time the reconcile was triggered.
And then compare `currentTime > reconcile-trigger-date`

schrodit: ah I just saw that the actual time is set.
So I would just rename the annotation to reflect what is actually is.
Something like `reconcile-time` or `reconcile-trigger-time`

pkg/deployer/container/reconciler_deployitem.go
```diff
@@ -66,8 +68,13 @@ func (r *DeployItemReconciler) Reconcile(ctx context.Context, req reconcile.Requ
...
 	}
 
-	if deployItem.Status.ObservedGeneration == deployItem.Generation && !lsv1alpha1helper.HasOperation(deployItem.ObjectMeta, lsv1alpha1.ReconcileOperation) {
-		logger.V(5).Info("Version already reconciled")
+	err = dutils.HandleAnnotationsAndGeneration(ctx, r.lsClient, r.log, deployItem)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if deployItem.Status.Phase != lsv1alpha1.ExecutionPhaseInit && deployItem.Status.Phase != lsv1alpha1.ExecutionPhaseDeleting {
```
schrodit, line 76: this seems much cleaner now:)

pkg/deployer/mock/actuator.go
```diff
@@ -74,7 +76,13 @@ func (a *actuator) Reconcile(ctx context.Context, req reconcile.Request) (reconc
 		return reconcile.Result{}, nil
 	}
 
-	if deployItem.Status.ObservedGeneration == deployItem.Generation {
+	err := dutils.HandleAnnotationsAndGeneration(ctx, a.c, a.log, deployItem)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if deployItem.Status.Phase != lsv1alpha1.ExecutionPhaseInit && deployItem.Status.Phase != lsv1alpha1.ExecutionPhaseDeleting {
```
schrodit, line 84: shouldn't the mock deployer be able to always reconcile?
we need this check in the container deployer as it otherweise would deploy multiple pods but the other controllers use the reconcile also for generic reconciles

pkg/deployer/utils/utils.go
```diff
@@ -0,0 +1,60 @@
...
+	lsv1alpha1helper "github.com/gardener/landscaper/apis/core/v1alpha1/helper"
+)
+
+// HandleAnnotationsAndGeneration is meant to be called at the beginning of a deployer's reconcile loop.
+// If a reconcile is needed due to the reconcile annotation or a change in the generation, it will set the phase to Init and remove the reconcile annotation.
+// It will also remove the timeout annotation if it is set.
+// Returns:
+//   - the modified deployitem
+//   - an error, if updating the deployitem failed, nil otherwise
+func HandleAnnotationsAndGeneration(ctx context.Context, c client.Client, log logr.Logger, di *lsv1alpha1.DeployItem) error {
```
schrodit, line 23:
```suggestion
func HandleAnnotationsAndGeneration(ctx context.Context, log logr.Logger, c client.Client, di *lsv1alpha1.DeployItem) error {
```

pkg/deployer/utils/utils.go
```diff
@@ -0,0 +1,60 @@
...
+	changedMeta := false
+	changedStatus := false
+	hasReconcileAnnotation := lsv1alpha1helper.HasOperation(di.ObjectMeta, lsv1alpha1.ReconcileOperation)
+	old := di.DeepCopy()
+	if hasReconcileAnnotation || di.Status.ObservedGeneration != di.Generation {
+		// reconcile necessary
+		log.V(5).Info("reconcile required, setting observed generation and phase", "type", "DeployItem", "name", di.Name, "namespace", di.Namespace, "reconcileAnnotation", hasReconcileAnnotation, "observedGeneration", di.Status.ObservedGeneration, "generation", di.Generation)
```
schrodit, line 30: the name and namesapce is duplicated. I suggest to declare a specific logger at the beginning:
```
logger := log.WithValues("name", di.name, "namesapce", di.namespace)
logger.V(5).Info(...)
```

pkg/deployer/utils/utils.go
```diff
@@ -0,0 +1,60 @@
...
+
+// HandleAnnotationsAndGeneration is meant to be called at the beginning of a deployer's reconcile loop.
+// If a reconcile is needed due to the reconcile annotation or a change in the generation, it will set the phase to Init and remove the reconcile annotation.
+// It will also remove the timeout annotation if it is set.
+// Returns:
+//   - the modified deployitem
+//   - an error, if updating the deployitem failed, nil otherwise
+func HandleAnnotationsAndGeneration(ctx context.Context, c client.Client, log logr.Logger, di *lsv1alpha1.DeployItem) error {
+	changedMeta := false
+	changedStatus := false
```
schrodit, line 25: these 2 variables seems unnecessary.
Why not just update the status after line 37 and the item after line 42?

pkg/landscaper/controllers/deployitem/actuator.go
```diff
@@ -0,0 +1,126 @@
...
+func (a *actuator) InjectLogger(log logr.Logger) error {
+	a.log = log
+	return nil
+}
+
+// InjectScheme injects the current scheme into the actuator
+func (a *actuator) InjectScheme(scheme *runtime.Scheme) error {
+	a.scheme = scheme
+	return nil
+}
```
schrodit, line 63: inject methods are deprecated by the controller library we should not use them in the new controllers.
instead just add them during the initialization of the controller

pkg/landscaper/controllers/deployitem/actuator.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (a *actuator) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	a.log.Info("reconcile", "resource", req.NamespacedName)
```
schrodit, line 66:
```suggestion
	a.log.V(5).Info("reconcile", "resource", req.NamespacedName.String())
```

pkg/landscaper/deployitem/deployitem.go
```diff
@@ -0,0 +1,46 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
+	lsv1alpha1helper "github.com/gardener/landscaper/apis/core/v1alpha1/helper"
+	"github.com/gardener/landscaper/pkg/landscaper/operation"
+)
+
+// Operation contains all deployitem operations
+type Operation struct {
+	operation.Interface
```
schrodit, line 19: do you really need the whole operation interface?

pkg/landscaper/deployitem/deployitem.go
```diff
@@ -0,0 +1,46 @@
...
+// NewOperation creates a new deployitem operations
+func NewOperation(op operation.Interface, di *lsv1alpha1.DeployItem) *Operation {
+	return &Operation{
+		Interface: op,
+		di:        di,
+	}
+}
+
+// FailDueToTimeout sets the phase of the deployitem to failed and sets lastError to a timeout error
+func (o *Operation) FailDueToTimeout(ctx context.Context, timeoutOperation, timeoutReason, timeoutMessage string) error {
```
schrodit, line 32: the `timeoutOperation` and `timeoutReason` can be hardcoded in this package

schrodit (changes requested):
looks overall pretty good and clear.
As would like to have integration tests for that behavior as well as a documentation for the behavior and the deployer <-> landsacper contract would be great

apis/core/types_deployitem.go
Diaphteiros: Hm, I didn't think of this, but it sounds reasonable. Will implement it this way.

pkg/deployer/mock/actuator.go
Diaphteiros: Actually, most deployers had some kind of check `deployitem.Status.ObservedGeneration == deployitem.Generation` there which I replaced with the new logic. Not sure about the mock deployer, but I can remove it here.

pkg/deployer/utils/utils.go
Diaphteiros: Hm, we want to do as few updates as possible, because each one potentially causes conflicts with the deployers. The problem is that within the first `if` block, the metadata might or might not change, so we don't know for sure whether we need to update the object. If we update it directly after line 32, we might have to do it again if we go into the second `if` block, so we would have to update the object twice, which we are trying to avoid.

But I guess you are right about the status, this is only and always updated in the first `if`, so we could get rid of the `changedStatus` variable and do the update and the end of the first if.

WDYT?

pkg/landscaper/deployitem/deployitem.go
Diaphteiros: Probably not, but I thought it would make it easier to understand if this was handled the same way as in the other reconcile loops. Also, this way it is easier to extend if we need some additional modifications in future. But I can also get rid of it.
WDYT?

schrodit: as you only need the client I think there is no benefit using the whole operation.
If we need other methods in the future we can reconsider using the whole installation operation.

apis/config/v1alpha1/types_landscaper_config.go
```diff
@@ -25,6 +25,11 @@ type LandscaperConfiguration struct {
 	// CrdManagement configures whether the landscaper controller should deploy the CRDs it needs into the cluster
 	// +optional
 	CrdManagement *CrdManagementConfiguration `json:"crdManagement,omitempty"`
+	// DeployItemPickupTimeout defines how long a deployer can take to react on changes to a deploy item before the landscaper will mark it as failed.
+	// Time is specified in seconds. Set to -1 to disable timeout checking.
+	// Will be defaulted to 300 (5 minutes) if not specified.
+	// +optional
+	DeployItemPickupTimeout *int64 `json:"deployItemPickupTimeout,omitempty"`
```
schrodit, line 32: how about using a string and then parse it with `time.ParseDuration("1s")`.
With that we could use more human readable values. WDYT?

apis/core/v1alpha1/constants.go
```diff
@@ -19,6 +19,9 @@ const (
 	// OperationAnnotation is the annotation that specifies a operation for a component
 	OperationAnnotation = "landscaper.gardener.cloud/operation"
 
+	// TimeoutAnnotation is used to recognize timeouts in deployitems
+	TimeoutAnnotation = "landscaper.gardener.cloud/reconcile-time"
```
schrodit, line 23:
```suggestion
	ReconcileTimeAnnotation = "landscaper.gardener.cloud/reconcile-time"
```

pkg/landscaper/controllers/deployitem/actuator.go
```diff
@@ -0,0 +1,97 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
+	lsv1alpha1helper "github.com/gardener/landscaper/apis/core/v1alpha1/helper"
+	"github.com/gardener/landscaper/pkg/landscaper/deployitem"
+	"github.com/gardener/landscaper/pkg/landscaper/operation"
+)
+
+func NewActuator(log logr.Logger, c client.Client, scheme *runtime.Scheme, pickupTimeout *int64) (reconcile.Reconciler, error) {
```
schrodit, line 24: can you add a doc string for the actuator and explain a bit what the controller is doing.

schrodit (changes requested):

apis/config/v1alpha1/types_landscaper_config.go
Diaphteiros: Done :+1:

pkg/landscaper/controllers/deployitem/actuator.go
```diff
@@ -0,0 +1,102 @@
...
+	lsv1alpha1helper "github.com/gardener/landscaper/apis/core/v1alpha1/helper"
+	"github.com/gardener/landscaper/pkg/landscaper/deployitem"
+)
+
+// NewActuator creates a new deploy item controller that handles timeouts
+// To detect pickup timeouts (when a DeployItem resource is not reconciled by any deployer within a specified timeframe), the controller checks for a timestamp annotation.
+// It is expected that deployers remove the timestamp annotation from deploy items during reconciliation. If the timestamp annotation exists and is older than a specified duration,
+// the controller marks the deploy item as failed.
+// rawPickupTimeout is a string containing the pickup timeout duration, either as 'none' or as a duration that can be parsed by time.ParseDuration.
+func NewActuator(log logr.Logger, c client.Client, scheme *runtime.Scheme, rawPickupTimeout string) (reconcile.Reconciler, error) {
```
schrodit, line 28:
```suggestion
func NewController(log logr.Logger, c client.Client, scheme *runtime.Scheme, rawPickupTimeout string) (reconcile.Reconciler, error) {
```

pkg/landscaper/controllers/deployitem/actuator.go
```diff
@@ -0,0 +1,102 @@
...
+	timeoutLog := ""
+	if act.pickupTimeout != nil {
+		timeoutLog = act.pickupTimeout.String()
+	}
+	log.Info("deploy item pickup timeout detection", "active", act.pickupTimeout != nil, "timeout", timeoutLog)
+
+	return &act, nil
+}
+
+type actuator struct {
```
schrodit, line 53:
```suggestion
type controller struct {
```

pkg/landscaper/controllers/deployitem/actuator.go
```diff
@@ -0,0 +1,102 @@
...
+	log           logr.Logger
+	c             client.Client
+	scheme        *runtime.Scheme
+	pickupTimeout *time.Duration
+}
+
+func (a *actuator) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	logger := a.log.WithValues("resource", req.NamespacedName.String())
```
schrodit, line 61:
```suggestion
        if a.pickupTimeout == nil {
          return nil
        }
	logger := a.log.WithValues("resource", req.NamespacedName.String())
```
a not set pickup timeout should be handled

pkg/landscaper/deployitem/deployitem.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const (
+	DefaultPickupTimeout   = 5 * time.Minute    // default pickup timeout
+	PickupTimeoutReason    = "PickupTimeout"    // for error messages
+	PickupTimeoutOperation = "WaitingForPickup" // for error messages
+)
+
+// FailDueToTimeout sets the phase of the deployitem to failed and sets lastError to a timeout error
+func FailDueToTimeout(ctx context.Context, c client.Client, log logr.Logger, di *lsv1alpha1.DeployItem, timeoutMessage string) error {
```
schrodit, line 25: I think for simplicity this function can be directly inline in the reconicle function.
so no need for another package

test/integration/deployitems/pickup_timeout.go
```diff
@@ -0,0 +1,158 @@
...
+			// checking whether the set timestamp is up-to-date is difficult due to potential differences between the
+			// system times of the machine running the landscaper and the machine running the tests
+			// so just check for existence of the annotation
+			g.Expect(lsv1alpha1helper.HasReconcileTimestampAnnotation(di.ObjectMeta)).To(g.BeTrue())
+
+			ginkgo.By("check for pickup timeout")
+			time.Sleep(deployItemPickupTimeout)
+			success := false
+			timeoutTime := time.Now().Add(waitingForFailedState)
+			for { // wait for the annotation
```
schrodit, line 111: you could use ginkgos `Eventually` function

schrodit (changes requested):
I really like the deploy item documentation 👍🏻

apis/core/types_shared.go
```diff
@@ -106,6 +106,8 @@ const (
 	ErrorCleanupResources ErrorCode = "ERR_CLEANUP"
 	// ErrorConfigurationProblem indicates that the last error occurred due a configuration problem.
 	ErrorConfigurationProblem ErrorCode = "ERR_CONFIGURATION_PROBLEM"
+	// ErrorTimeout indicates that an operation timed out.
+	ErrorTimeout ErrorCode = "ERR_TIMEOUT"
```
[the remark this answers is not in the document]
Diaphteiros: I think it can be 'unretryable', but it also can be resolved by retrying, depending on the situation. E.g. if the deployer is currently unavailable (due to node rolling, ...) retrying at a later time could help. If the deployer isn't there, however, no amount of retries will solve the problem.
For other timeouts, such as the progressing timeout for example, retrying might have an even higher chance of solving the problem.
I think we should leave it retryable for now ... WDYT?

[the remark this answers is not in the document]
schrodit: You're basically right. But currently we have the issue that the behavior for installations that were in failed state and switch to successfull without being triggered is undefined.

So I would make it not retryable and then later when we have proper propagation for such scenarios we will remove it again and make it retryable.

docs/technical/deployer_contract.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Status
+
+Once handled by its deployer, a status similar to this one will be attached to the deploy item:
+```yaml
+status:
+  observedGeneration: 1
+  phase: Succeeded
+  providerStatus:
+    apiVersion: manifest.deployer.landscaper.gardener.cloud/__internal
```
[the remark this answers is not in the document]
Diaphteiros: I just checked and it seems that only the manifest deployer adds the internal status, for the other deploy items, it is versioned.

pkg/landscaper/controllers/deployitem/actuator.go
Diaphteiros: It is handled, but in a different way:
It is possible to configure the landscaper to not do any pickup timeout checks by setting the timeout in the config to `none`. In this case, `a.pickupTimeout` will be set to `nil`, otherwise it will contain a `time.Duration`.

Diaphteiros:
@schrodit I addressed all of your suggestions, except the internal version for the deploy item status - I fixed that in the documentation, but the fix in the code doesn't belong in this PR, so I created an issue: #152

schrodit (changes requested):
just some minor styling things rest is lgtm.
You also have to rebase after the recent PR merges

schrodit (approved):
/lgtm

PR #145, opened 2021-03-17
Fix a nil pointer when a configmap reference data import key does not exist

schrodit:

/kind bug
/priority 3

Fixes a bug in the installation data import that paniced when a key of a configmap was not defined.

also added some unit tests for the dataobject and secret import

Fixes #

**Release note**:

```bugfix user
Fixed a bug in the Landscaper controller that caused a panic when a imported configmap key was undefined.
```


Diaphteiros (approved):
/lgtm

PR #146, opened 2021-03-18
create own binary for webhook server

schrodit:

/area high-availability
/kind enhancement
/priority 3

The validation webhooks and landscaper controller are now split into different containers.
This is done to ensure more stability for the webhooks which can be easily scaled.
Now the controller can work fully asynchronisly and can be restarted, rescheduled at any time without downtime of the validation.

IN addition, the certificate creation has been improved to also update the certificate on changes in the dns names.

Fixes #

The intergration tests are run locally as with the new image that needs to be build the testmachinery tests will fail

**Release note**:

```feature operator
The validation webhook and the landscaper controller are now split into different containers and deployment.
This is done to ensure isolation of validation and the actual load in the controller.

Checkout the updated landscaper helm chart.
```


charts/landscaper/templates/service.yaml
```diff
@@ -16,11 +16,23 @@ spec:
...
+kind: Service
+metadata:
+  name: {{ include "landscaper.webhooks.fullname" . }}
+  labels:
+  {{- include "landscaper.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+  - port: {{ .Values.webhookConfig.servicePort }}
+    targetPort: {{ .Values.webhookConfig.servicePort }}
```
Diaphteiros, line 33: I might be confused here, but I think the `targetPort` has to match the webserver's port, which is hardcoded to `9443` [here](https://github.com/schrodit/landscaper/blob/standalone-webhooks/cmd/landscaper-webhooks-server/app/app.go#L57).

Diaphteiros (changes requested):
I'm not completely sure, but the port thing looks wrong to me. Apart from that it looks good.

Makefile
```diff
@@ -45,7 +46,7 @@ test:
 
 .PHONY: integration-test
 integration-test:
-	@go test -mod=vendor $(REPO_ROOT)/test/integration --v -ginkgo.v -ginkgo.progress --kubeconfig $(KUBECONFIG) --ls-version $(EFFECTIVE_VERSION) --registry-config=$(REGISTRY_CONFIG)
+	@go test -timeout=0 -mod=vendor $(REPO_ROOT)/test/integration --v -ginkgo.v -ginkgo.progress --kubeconfig $(KUBECONFIG) --ls-version $(EFFECTIVE_VERSION) --registry-config=$(REGISTRY_CONFIG)
```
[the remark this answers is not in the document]
schrodit: this make command is only used locally. we have a dedicated testdefinition with its own timeout of 30min for the pipeline executions.
The idea is that locally we do not need a timeout as there is someone watching/executing the test and can anytime abort the test with ctrl+c.
But I think we could also set a timeout to 1h or so

charts/landscaper/templates/service.yaml
schrodit: oh yes good catch. I think I will make it configurable. (the idea is that someone running it locally can run the landsacper controller and the webhook server locally)

Makefile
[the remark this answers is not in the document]
Diaphteiros: Ah, no, I didn't know that the pipeline has the timeout implement in another way. So this is fine for me.

cmd/landscaper-webhooks-server/app/options.go
```diff
@@ -0,0 +1,168 @@
...
+		if len(ws) < 2 {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("--webhook-service"), o.webhookServiceNamespaceName, "must have the format '<namespace>/<name>'"))
+		} else {
+			if len(ws[0]) == 0 || len(ws[1]) == 0 {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("--webhook-service"), o.webhookServiceNamespaceName, "neither name nor namespace of the webhook service must be empty"))
+			}
+		}
+	}
+	// validate port
+	if o.webhookServicePort <= 0 {
```
Diaphteiros, line 127: Should we add a check like this also for `o.port`?

Diaphteiros (approved):
/lgtm

PR #151, opened 2021-03-29
fix conversion with internal versions

schrodit:
also renamed kubernetes package to more suitable api package

/area robustness
/kind bug
/priority 3

Currently we are unable to convert api versions through their internal type using the default codec factory decoder.
This PR adds our own decoding logic that automatically uses the internal type to perform conversion.

Fixes the manifestv1alpha1 to manifestvalpha2 issue.

Also fixed abug in the mock deployer that still had used the helm scheme (no idea how that could have ever worked...).

Fixes #

I also moved the landsacper api schema from `pkg/kubernetes` to `pkg/api` which I think is the more natural place for such variables.

**Release note**:

```bugfix developer
A bug that prevents the conversion from manifest v1alpha1 to v1alpha2 has been fixed.
```
```breaking developer
The Landscaper specific schemes and decoders have been move from `pkg/kubernetes` to `pkg/apis`. Please update your imports accordingly.
```


MrBatschner (approved):
/LGTM

PR #157, opened 2021-04-01
Deploy Item Timeout Detection

Diaphteiros:

/area robustness
/kind enhancement
/priority 3

This PR builds upon #143 which added detection and a timeout for when a deploy item is not handled by any deployer.
This is now expanded:
If a deploy item is in a non-final phase for too long, which most probably means that it is stuck in `Progressing` for some reason, the landscaper will annotate it with the abort operation annotation `landscaper.gardener.cloud/operation: abort`, instructing the corresponding deployer to gracefully stop processing this deploy item.
Together with the abort operation annotation, there is also a timestamp set at the deploy item, which is used by the landscaper to monitor how long the deployer takes to 'abort' processing the deploy item. If this takes too long, the landscaper will set the deploy item to `Failed` and add a timeout error to its status.

This branch is based on the one from #143, so until that one is merged, the GitHub diff will show its changes too.

**Release note**:

```other operator
TODO
```


Diaphteiros:
replaced by #174

PR #159, opened 2021-04-07
Integration test logging to include start/end timestamps

MrBatschner:

/area testing
/kind enhancement
/priority 4

This PR changes the log collection in the integration test framework not to collect every log taken since the beginning of the all tests but only since a specific timestamp which will be set to the start-time of each individual test for each individual test. It makes reading the logs much easier because it eliminates duplicated output.

**Release note**:

```other developer
Elimination of duplication of logs from integration tests.
```

### DIFF
--- a/test/framework/dump.go
+++ b/test/framework/dump.go
@@ -10,10 +10,12 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


schrodit (approved):
/lgtm

PR #164, opened 2021-04-08
update component-cli lib and simplify ctf build

schrodit:

/area delivery
/priority 3

Simplify the landscaper artifact generation within the component descriptor generation

also added a local script to test the ctf generation

Fixes #

**Release note**:

```other operator
NONE
```


hack/generate-cd.sh
```diff
@@ -0,0 +1,61 @@
...
+
+if ! which component-cli 1>/dev/null; then
+  echo -n "component-cli is required to generate the component descriptors"
+  echo -n "Trying to installing it..."
+  go get github.com/gardener/component-cli/cmd/component-cli
+
+  if ! which component-cli 1>/dev/null; then
+    echo -n "component-cli was successfully installed but the binary cannot be found"
+    echo -n "Try adding the \$GOPATH/bin to your \$PATH..."
+    go get github.com/gardener/component-cli/cmd/component-cli
```
MrBatschner, line 27: If $GOPATH is not in $PATH, running `go get` a second time probably will not help, instead, we should exit.

hack/generate-cd.sh
```diff
@@ -0,0 +1,61 @@
...
+#access:
+#  type: 'ociRegistry'
+#  imageReference: ${image}
+#EOF
+#done
+
+echo "> Creating ctf"
+CTF_PATH=./gen/ctf.tar
+mkdir -p ./gen
+rm ${CTF_PATH}
```
MrBatschner, line 60: if `./gen/ctf.tar` did not exist and with `set -e` set, the script will exit here
```suggestion
[ -e $CTF_PATH ] && rm ${CTF_PATH}
```

MrBatschner (commented):
Just two minor things... and something does not seem to be ok with building the component-descriptors, it fails in the publish step - see below.

.ci/component_descriptor
```diff
@@ -23,21 +23,12 @@ function buildComponentArchive() {
   CA_PATH="$(mktemp -d)"
   printf "> Building component ${COMPONENT_NAME}\n"
 
-  component-cli ca resources add ${CA_PATH} \
+  component-cli ca "${CA_PATH}" "${CTF_PATH}" \
```
MrBatschner, line 26:
```suggestion
  component-cli ca create "${CA_PATH}" "${CTF_PATH}" \
```

schrodit: this is right. but the we need to wait for the new component-cli version to the reach concourse.

PR #167, opened 2021-04-09
add health helper functions and restructure utils

schrodit:

/kind cleanup
/priority 3

adds a dedicated health helper package to the v1alpha1 apis.
This is mostly copied from https://github.com/gardener/gardener/pull/3757/files so credit goes to @danielfoehrKn .

Also restructured the other helper functions in the landscaper code.

@danielfoehrKn I also tried to add the waiting function to the apis package but would have introduced a dependency towards the controller-runtime which we should avoid.
But at least the health helper could be shared

**Release note**:

```other developer
A dedicated health helper has been introduced that can be used to determine if a Installation is healhty.
```

### DIFF
--- a/apis/core/v1alpha1/health/health.go
+++ b/apis/core/v1alpha1/health/health.go
@@ -1,0 +1,85 @@
+// SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package health
+
+import (
+	"fmt"
+
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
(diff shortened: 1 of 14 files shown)


apis/core/v1alpha1/health/health.go
```diff
@@ -0,0 +1,85 @@
...
+		lsv1alpha1.ValidateImportsCondition,
+	}
+)
+
+// CheckInstallation checks if the given Installation is healthy
+// An installation is healthy if
+// * Its observed generation is up-to-date
+// * No annotation landscaper.gardener.cloud/operation is set
+// * No lastError is in the status
+// * A last operation is state succeeded is present
```
MrBatschner, line 34:
```suggestion
// * A last operation in state succeeded is present
```

MrBatschner (commented):
One typo in one comment. ;-)
Apart from that,
/lgtm

PR #170, opened 2021-04-09
depend on publish-trait, rather than on injected steps

ccwienk:
prepare switching to kaniko as default-builder


schrodit (approved):
/lgtm

PR #172, opened 2021-04-13
make webhookserver and controller image tags configurable

schrodit:

/area delivery
/priority 3

It is now possible to configure the webhook server image and controller image separately.
the old `image.tag` is still valid.

Fixes #

cc @achimweigel @robertgraeff 

**Release note**:

```other operator
It is now possible to configure the webhook server image and controller image separately.
```


achimweigel (commented):
The overall reconcile logic looks good and resolves the problem that reconcile annotations got lost if subsequent status updates fails.

PR #174, opened 2021-04-14
Deploy Item Timeout Detection

### LINKED ISSUE #115: DeployItem Timeout

Diaphteiros:

/area robustness
/kind enhancement
/priority 3

This PR builds upon #143 which added detection and a timeout for when a deploy item is not handled by any deployer.
This is now expanded:
If a deploy item is in a non-final phase for too long, which most probably means that it is stuck in `Progressing` for some reason, the landscaper will annotate it with the abort operation annotation `landscaper.gardener.cloud/operation: abort`, instructing the corresponding deployer to gracefully stop processing this deploy item.
Together with the abort operation annotation, there is also a timestamp set at the deploy item, which is used by the landscaper to monitor how long the deployer takes to 'abort' processing the deploy item. If this takes too long, the landscaper will set the deploy item to `Failed` and add a timeout error to its status.

Fixes #115 

This branch is based on the one from #143, so until that one is merged, the GitHub diff will show its changes too.

**Release note**:

```feature operator
The landscaper now detects if a deploy item is in a non-final state for too long, which is interpreted as the deployer getting stuck while processing it. It will then get the abort operation annotation, giving the deployer the opportunity to abort processing the deploy item gracefully. If that also times out, landscaper will set the deploy item to `Failed`. Have a look at the [documentation](docs/technical/deployer_contract.md) for further information on the timeouts.
```


apis/config/types_landscaper_config.go
```diff
@@ -25,11 +25,28 @@ type LandscaperConfiguration struct {
...
 	// Allowed values are 'none' (to disable pickup timeout detection) and anything that is understood by golang's time.ParseDuration method.
 	// Defaults to five minutes if not specified.
 	// +optional
-	DeployItemPickupTimeout string `json:"deployItemPickupTimeout,omitempty"`
+	Pickup string `json:"pickup,omitempty"`
+	// Abort specifies how long the deployer may take to abort handling a deploy item after getting the abort annotation.
+	// Allowed values are 'none' (to disable abort timeout detection) and anything that is understood by golang's time.ParseDuration method.
+	// Defaults to five minutes if not specified.
+	// +optional
+	Abort string `json:"abort,omitempty"`
```
schrodit, line 44: as i currently see the string the copied documentation of allowed values.
I would propose to do something like kubernetes did with the `metav1.Time`.
So basically lets create our own duration wrapper that implements the json marshal/unmarshal methods + the opeapispec values.
With that we can simply write:
```
Abort *Duration `json:"abort,omitempty"`
```
So we do not need special parse logic in the business logic and the api seems clearer as we actually want a duration rather than a string.

apis/core/v1alpha1/helper/helpers.go
```diff
@@ -32,24 +39,29 @@ func SetOperation(obj *metav1.ObjectMeta, op v1alpha1.Operation) {
...
 }
 
-// HasReconcileTimestampAnnotation checks if the obj has the given timeout annotation
-func HasReconcileTimestampAnnotation(obj metav1.ObjectMeta) bool {
-	_, ok := obj.Annotations[v1alpha1.ReconcileTimestampAnnotation]
+// HasTimestampAnnotation checks if the obj has the given timeout annotation
+func HasTimestampAnnotation(obj metav1.ObjectMeta, ta TimestampAnnotation) bool {
+	_, ok := obj.Annotations[string(ta)]
 	return ok
 }
```
schrodit, line 46: not sure if we need that function.
As you already pass a annotation there is no difference to the default `HasAnnotation` function?
(when calling the default function e.g. metav1.HasAnnotation(obj, v1alpha1.ReconcileTimestampAnnotation), you also directly see that you want the reconcile timestamp)

apis/core/v1alpha1/helper/helpers.go
```diff
@@ -32,24 +39,29 @@ func SetOperation(obj *metav1.ObjectMeta, op v1alpha1.Operation) {
...
+	SetTimestampAnnotation(obj, ta, time.Now())
 }
 
-// SetReconcileTimestampAnnotationNow sets the timeout annotation with the current timestamp.
-func SetReconcileTimestampAnnotationNow(obj *metav1.ObjectMeta) {
-	SetReconcileTimestampAnnotation(obj, time.Now())
+// SetTimestampAnnotation sets the timeout annotation with the given timestamp.
+func SetTimestampAnnotation(obj *metav1.ObjectMeta, ta TimestampAnnotation, ts time.Time) {
+	metav1.SetMetaDataAnnotation(obj, string(ta), ts.Format(time.RFC3339))
 }
```
schrodit, line 60: this function is only used in one package and it does abstract that much.
I would suggest to either remove the helper and just call directly `metav1.SetMetaDataAnnotation(obj, string(ta), ts.Format(time.RFC3339))` or move the function(s) to the respective package

apis/deployer/mock/types_provider.go
```diff
@@ -22,6 +22,10 @@ type ProviderConfiguration struct {
 	// Phase sets the phase of the DeployItem
 	Phase *lsv1alpha1.ExecutionPhase `json:"phase,omitempty"`
 
+	// InitialPhase sets the phase of the DeployItem, but only if it is empty or "Init"
+	// Additionally, setting it will suppress the DeployItem phase being set to "Succeeded" after successful reconciliation
+	InitialPhase *lsv1alpha1.ExecutionPhase `json:"initialPhase,omitempty"`
```
schrodit, line 27: why is the default phase not enough?

docs/technical/deployer_contract.md
```diff
@@ -103,15 +104,22 @@ As explained above, even if the type is correct, the deployer might still not be
 
 #### 3. Handle Annotations
 There are two important annotations that need to be handled by the deployer: 
-The reconcile annotation `landscaper.gardener.cloud/operation: reconcile` indicates that either a human operator or the landscaper wants this deploy item to be reconciled. The deployer has to remove this annotation. In addition, it should set the deploy item's phase to `Init` to show the beginning of a new reconciliation and avoid loss of information in case the deployer dies immediately after removing the annotation.
+The operation annotation `landscaper.gardener.cloud/operation: reconcile` indicates that either a human operator or the landscaper wants a specific operation to be fulfilled on this deploy item. The value of the annotation specifies the expected operation:
```
schrodit, line 107: very well written doc :)

pkg/landscaper/controllers/deployitem/add.go
```diff
@@ -11,15 +11,8 @@ import (
...
-// It is responsible for detecting timeouts in deploy items.
-func AddControllerToManager(mgr manager.Manager, rawDeployItemPickupTimeout string) error {
-	if rawDeployItemPickupTimeout == "none" {
-		// currently the deploy item reconcile loop is only used for pickup timeout detection
-		// so if that is disabled there is no need to watch deploy items
-		return nil
-	}
-	a, err := NewController(ctrl.Log.WithName("controllers").WithName("DeployItem"), mgr.GetClient(), mgr.GetScheme(), rawDeployItemPickupTimeout)
+func AddControllerToManager(mgr manager.Manager, rawDeployItemPickupTimeout, rawDeployItemAbortingTimeout, rawDeployItemDefaultTimeout string) error {
+	a, err := NewController(ctrl.Log.WithName("controllers").WithName("DeployItem"), mgr.GetClient(), mgr.GetScheme(), rawDeployItemPickupTimeout, rawDeployItemAbortingTimeout, rawDeployItemDefaultTimeout)
```
schrodit, line 15: nit
```suggestion
	a, err := NewController(ctrl.Log.WithName("controllers").WithName("DeployItem"), 
	  mgr.GetClient(), 
	  mgr.GetScheme(), 
	  rawDeployItemPickupTimeout, 
	  rawDeployItemAbortingTimeout, 
	  rawDeployItemDefaultTimeout,
	)
```

pkg/landscaper/controllers/deployitem/controller.go
```diff
@@ -40,30 +43,54 @@ func NewController(log logr.Logger, c client.Client, scheme *runtime.Scheme, raw
...
+	}
+	if rawDefaultTimeout == "none" {
+		con.defaultTimeout = nil
+	} else {
+		tmp, err := time.ParseDuration(rawDefaultTimeout)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse deploy item default timeout into a duration: %w", err)
+		}
+		con.defaultTimeout = &tmp
+	}
```
schrodit, line 63: if we use my propsed change to the api. this code will become unnecessary and way less complex

pkg/landscaper/controllers/deployitem/controller.go
```diff
@@ -75,35 +102,175 @@ func (con *controller) Reconcile(ctx context.Context, req reconcile.Request) (re
 		return reconcile.Result{}, err
 	}
 
+	var requeue *time.Duration
+	var err error
+	old := di.DeepCopy()
+
+	// detect pickup timeout
+	if con.pickupTimeout != nil {
+		logger.V(5).Info("check for pickup timeout")
```
schrodit, line 111:
```suggestion
		logger.V(7).Info("check for pickup timeout")
```

pkg/landscaper/controllers/deployitem/controller.go
```diff
@@ -75,35 +102,175 @@ func (con *controller) Reconcile(ctx context.Context, req reconcile.Request) (re
...
+		}
+	}
+	if !reflect.DeepEqual(old.Status, di.Status) {
+		if err := con.c.Status().Update(ctx, di); err != nil {
+			logger.Error(err, "unable to set deployitem status")
+			return reconcile.Result{}, err
+		}
+		// if there was a pickup timeout, no need to check for anything else
+		return reconcile.Result{}, nil
+	}
```
schrodit, line 124: this check can be also move inside the pickup timeout if

pkg/landscaper/controllers/deployitem/controller.go
```diff
@@ -75,35 +102,175 @@ func (con *controller) Reconcile(ctx context.Context, req reconcile.Request) (re
...
+		}
+	}
+	if !reflect.DeepEqual(old.Status, di.Status) {
+		if err := con.c.Status().Update(ctx, di); err != nil {
+			logger.Error(err, "unable to set deployitem status")
+			return reconcile.Result{}, err
+		}
+		// if there was an aborting timeout, no need to check for anything else
+		return reconcile.Result{}, nil
+	}
```
schrodit, line 146: can also be move inside the above if block

test/integration/deployitems/aborting_timeout.go
```diff
@@ -0,0 +1,82 @@
+// SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors.
```
schrodit, line 1: I think this test does not need a real k8s cluster.
We could write it as unit test and the envtest lib

test/integration/deployitems/common.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	waitingForDeployItems     = 5 * time.Second  // how long to wait for the landscaper to create deploy items from the installation
+	deployItemPickupTimeout   = 10 * time.Second // the landscaper has to be configured accordingly for this test to work!
+	deployItemAbortingTimeout = 10 * time.Second // the landscaper has to be configured accordingly for this test to work!
+	waitingForReconcile       = 10 * time.Second // how long to wait for the landscaper or the deployer to reconcile and update the deploy item
+	resyncTime                = 1 * time.Second  // after which time to check again if the condition was not fulfilled the last time
+	retryUpdates              = 5 * time.Second  // for how long update operations are retried if they failed
+)
+
+func namespacedName(meta metav1.ObjectMeta) types.NamespacedName {
```
schrodit, line 32: there is already a helper function in our `kutil` package

test/integration/deployitems/timeout.go
```diff
@@ -0,0 +1,107 @@
+// SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors.
```
schrodit, line 1: couldn't we test that also as unit test with the envtest?

schrodit (changes requested):

apis/deployer/mock/types_provider.go
Diaphteiros: Because the deployer sets the phase in each reconcile loop. This means I can't test whether the phase is `Failed` after it has been in `Progressing` for too long, because it will immediately become `Progressing` again. Therefore I need an option to set the phase to `Progressing`, but don't overwrite the `Failed` afterwards.

apis/core/v1alpha1/helper/helpers.go
Diaphteiros: I removed that one and only kept the `SetTimestampAnnotationNow` method - is this ok or would you like me to move it out of this package too?

test/integration/deployitems/aborting_timeout.go
Diaphteiros: I wrote unit tests for all timeouts and refactored the integration tests into one single (slightly bigger) test that verifies the timing and cooperation between the deploy item controller and the deployer.

test/integration/deployitems/timeout.go
Diaphteiros: see above

Diaphteiros:
@schrodit Please have a look again.

apis/core/v1alpha1/types_deployitem.go
```diff
@@ -83,6 +90,10 @@ type DeployItemStatus struct {
 	// LastError describes the last error that occurred.
 	LastError *Error `json:"lastError,omitempty"`
 
+	// LastChangeReconcileTime indicates when the reconciliation of the last change to the deploy item has started
+	// +optional
+	LastChangeReconcileTime *metav1.Time `json:"lastChangeReconcileTime,omitempty"`
```
schrodit, line 95: I think that the naming is a little bit long.
I think `LastReconcileTime` also describes what this field indicates. wdyt?

pkg/landscaper/controllers/deployitem/controller.go
```diff
@@ -15,55 +16,50 @@ import (
...
 )
 
 // NewController creates a new deploy item controller that handles timeouts
 // To detect pickup timeouts (when a DeployItem resource is not reconciled by any deployer within a specified timeframe), the controller checks for a timestamp annotation.
 // It is expected that deployers remove the timestamp annotation from deploy items during reconciliation. If the timestamp annotation exists and is older than a specified duration,
 // the controller marks the deploy item as failed.
-// rawPickupTimeout is a string containing the pickup timeout duration, either as 'none' or as a duration that can be parsed by time.ParseDuration.
-func NewController(log logr.Logger, c client.Client, scheme *runtime.Scheme, rawPickupTimeout string) (reconcile.Reconciler, error) {
+// pickupTimeout is a string containing the pickup timeout duration, either as 'none' or as a duration that can be parsed by time.ParseDuration.
+func NewController(log logr.Logger, c client.Client, scheme *runtime.Scheme, pickupTimeout, abortingTimeout, defaultTimeout lscore.Duration) (reconcile.Reconciler, error) {
```
schrodit, line 38: the code is totally fine but I think the dereference in the addController method is strange. and could lead to strange nil pointer exceptions

```suggestion
func NewController(log logr.Logger, c client.Client, scheme *runtime.Scheme, pickupTimeout, abortingTimeout, defaultTimeout *lscore.Duration) (reconcile.Reconciler, error) {
```

pkg/landscaper/controllers/deployitem/controller.go
```diff
@@ -75,35 +71,171 @@ func (con *controller) Reconcile(ctx context.Context, req reconcile.Request) (re
...
+			return reconcile.Result{}, err
+		}
+		if requeue == nil {
+			requeue = tmp
+		} else if tmp != nil && *tmp < *requeue {
+			requeue = tmp
+		}
+		if !reflect.DeepEqual(old.Status, di.Status) {
+			if err := con.c.Status().Update(ctx, di); err != nil {
+				logger.Error(err, "unable to set deployitem status")
```
schrodit, line 109: can you add a comment that we might have to expose such errors as events on the deploy item

schrodit (commented):
only some minor comments.
But overall lgtm.

Can also add our own duration type to the helm/manifest deployer timeout declaration so that we use the same types in our codebase to describe durations.

apis/core/v1alpha1/types_deployitem.go
Diaphteiros: The problem that I have with this is that we also use the term `reconcile` for each call to the deployers `Reconcile` method. We are 'reconciling' a deploy item a lot more than it actually changes, but this value is meant to contain the timestamp of the time when we started _reconciling the latest change to the deploy item_. That's why in my opinion `lastReconcileTime` is somewhat imprecise and I don't really like it. 
I get your point though, `lastChangeReconcileTime` is not really nice either. 

Maybe we can find some other term for this ... how about these:
- LastChangeReconciled (basically dropped the 'time' suffix)
- LastChanged (also imprecise, as it refers to the time the deployer has picked up the change and not the change itself ...)
- LastChangePickup

schrodit (approved):
/lgtm

PR #175, opened 2021-04-15
Fix initialisation with default context

achimweigel:

/area control-plane
/kind bug
/priority 3

Fixes problems with root installations with no repository context.

Assume Landscaper is configured with a default repository context. If there is a root installation without a repository context deployed, all subsequently deployed root installations will fail. The reason is that for processing new root installations, all already existing root installation are instantiated and this fails for those without a repository context.

**Release note**:

```other operator

```


schrodit (changes requested):
PR is lgtm but can you add a unit test checks that the default repository is correctly set

achimweigel:
we have added a unit test

schrodit (approved):
/lgtm

PR #191, opened 2021-05-06
Conditional imports

### LINKED ISSUE #61: Conditional imports

/area landscaper
/kind enhancement
/priority normal

The landscaper needs 'conditional imports' - imports that are required, if a certain other optional import is given, but are not evaluated if that other import is not specified.

**Why is this needed**:
For example the etcd backup: If activated, the landscaper needs iaas credentials and configuration for the backup blobstore. If not activated, this information is neither needed nor used. There are several more cases where some imports are only needed if another import signalizes that a specific option is activated (or similar cases).

Diaphteiros:

/area operations
/kind enhancement
/priority 3

This PR makes conditional imports work as expected.

Fixes #61 

**Release note**:

```feature user
It is now possible to define 'conditional imports' in blueprints. Conditional imports are imports which are only checked for if another import is satisfied. See the [documentation](docs/usage/ConditionalImports.md) for further information.
```

### DIFF
--- a/docs/usage/ConditionalImports.md
+++ b/docs/usage/ConditionalImports.md
@@ -1,0 +1,82 @@
+# Conditional Imports
+
+The basics of import definitions in blueprints are explained [here](./Blueprints.md). This document focuses on an advanced feature for import definitions: conditional imports.
+
+The idea is simple: sometimes some values are only needed if some others - which are optional - are also provided. You can also think about this the other way around, for example: if no url for a backup service is provided, we don't need any password for accessing it either. 
+
+Modelling this in a blueprint's import definitions is difficult - one can either mark the password as optional, but then landscaper can't enforce its existence, even if the url import is satisfied. Or they are marked as required, but then they need to be provided, even if they are not used.
+
+The concept of 'conditional imports' aims to solve this problem. The above example could look like this in a blueprint's imports definition:
+
+```yaml
+imports:
+- name: backup-storage-url
+  required: false
+  schema:
+    type: string
+  imports:
+  - name: backup-storage-password
+    schema:
+      type: string
+```
+
+Put in simple words, this definitions says "if the 'backup-storage-url' import is satisfied, the 'backup-storage-password' import has to be too".
+Please note that the `backup-storage-url` import is marked as optional. This always has to be the case and is validated by the landscaper.
+
+Nested 'conditional import' definitions follow exactly the same structure as top-level import definitions, which means they can also be marked as optional and contain further conditional imports depending on them.
+
+It is currently not possible to construct a condition that checks an existing import for a specific value - only whether an import is satisfied can be used for the condition.
+
+Unlike imports, exports can neither be made optional nor conditional.
+
+
+## Conditional Imports and Subinstallation Templates
+
+Sometimes a subinstallation has an import that comes from an optional or conditional import of the parent installation. This means that the subinstallation template in the parent's blueprint will have imports listed which might not be there during runtime.
+
+See the below snippet of a blueprint as an example:
+```yaml
+... # redacted
+imports:
+- name: foo
+  required: false
+  schema:
+    type: string
+  imports:
+  - name: bar
+    schema:
+      type: string
+
+subinstallations:
+- apiVersion: landscaper.gardener.cloud/v1alpha1
+  kind: InstallationTemplate
+  name: subinst-import
+  imports:
+    data:
+    - name: internalFoo
+      dataRef: foo
+    - name: internalBar
+      dataRef: bar
+  blueprint:
+    filesystem:
+      blueprint.yaml: |
+        ... # redacted
+        imports:
+        - name: internalFoo
+          required: false
+          schema:
+            type: string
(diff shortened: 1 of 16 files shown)


pkg/landscaper/installations/installation.go
```diff
@@ -48,14 +48,27 @@ func (i *Installation) ImportStatus() *ImportStatus {
 
 // GetImportDefinition return the import for a given key
 func (i *Installation) GetImportDefinition(key string) (lsv1alpha1.ImportDefinition, error) {
-	for _, def := range i.Blueprint.Info.Imports {
+	imports := getFlattenedImports(i.Blueprint.Info.Imports)
```
Diaphteiros, line 51: @schrodit I'd like to have your opinion on this:
This check is called to see whether the parent satisfies all imports of the subinstallation. Before my changes, this looped over `i.Blueprint.Info.Imports` (over all direct imports of the parent) and checked whether the referenced import exists. With conditional imports, this is not enough anymore.

What I changed: It now loops over a flattened import list, which contains not only the direct import definitions of the parent, but also the conditional ones - independently of whether the condition is fulfilled or not.

The potential problem I see: This doesn't check whether the condition for conditional imports is fulfilled. So, in theory, an import definition could be found which doesn't actually exist, as it depends on another import which is not fulfilled. In practise, this shouldn't happen, as subinstallation imports that reference conditional parent imports are dynamically removed from the subinstallation template, if the condition which would provide the parent import is not fulfilled.

However, this still feels pretty 'dirty' and I'm not sure that I didn't miss a case in which this assumption breaks.
Should we build it in a way that the flattened list only contains conditional import definitions, if the corresponding condition is fulfilled? If so, how do we get the information which conditional imports are fulfilled at this point (`i.Imports` is `nil`)?
What do you think?

docs/usage/ConditionalImports.md
```diff
@@ -0,0 +1,82 @@
+# Conditional Imports
```
schrodit, line 1: can you link that guide also in the `Bllueprint.md`

pkg/landscaper/installations/imports/conditional_imports_test.go
```diff
@@ -0,0 +1,192 @@
...
+			state *envtest.State
+		)
+
+		instRef = kutil.ObjectKey("conditional-import-inst", "test8")
+		cmRef = lsv1alpha1.ObjectReference{
+			Name:      "inst-import",
+			Namespace: instRef.Namespace,
+		}
+
+		fakeClient, state, err = envtest.NewFakeClientFromPath("./testdata/state")
```
schrodit, line 55: if the state is not needed you could also write
```suggestion
		fakeClient, _, err = envtest.NewFakeClientFromPath("./testdata/state")
```

pkg/landscaper/installations/imports/constructor.go
```diff
@@ -67,10 +80,20 @@ func (c *Constructor) Construct(ctx context.Context, inst *installations.Install
...
 					continue // don't throw an error if the import is not required
 				}
-				return installations.NewImportNotFoundErrorf(nil, "no import for %s exists", def.Name)
+				return nil, installations.NewImportNotFoundErrorf(nil, "no import for %s exists", def.Name)
 			}
 			if err := c.JSONSchemaValidator().ValidateGoStruct(def.Schema.RawMessage, imports[def.Name]); err != nil {
-				return installations.NewErrorf(installations.SchemaValidationFailed, err, "%s: imported datatype does not have the expected schema", defPath.String())
+				return imports, installations.NewErrorf(installations.SchemaValidationFailed, err, "%s: imported datatype does not have the expected schema", defPath.String())
+			}
+			if def.ConditionalImports != nil && len(def.ConditionalImports) > 0 {
```
schrodit, line 88: in go slices that are nil do also have a length of `0`
```suggestion
			if len(def.ConditionalImports) > 0 {
```

pkg/landscaper/installations/operation.go
```diff
@@ -465,6 +469,12 @@ func (o *Operation) CreateOrUpdateImports(ctx context.Context) error {
 			return fmt.Errorf("import %s not defined", importDef.Name)
 		}
 
+		if importDef.ConditionalImports != nil && len(importDef.ConditionalImports) > 0 {
```
schrodit, line 472: same as above
```suggestion
		if len(importDef.ConditionalImports) > 0 {
```

pkg/landscaper/installations/subinstallations/helper.go
```diff
@@ -130,6 +126,27 @@ func (o *Operation) ValidateSubinstallations(installationTmpl []*lsv1alpha1.Inst
 	return nil
 }
 
+func (o *Operation) buildCoreImports(importList lsv1alpha1.ImportDefinitionList) ([]core.ImportDefinition, error) {
```
schrodit, line 129: can you a dd a docstring explaining what the function is doing

schrodit (changes requested):

schrodit (approved):
/lgtm

PR #194, opened 2021-05-12
Not fetching the blueprint of installations if not needed

achimweigel:

/area control-plane
/kind bug
/priority 3

A root installation fails if the resolution of the blueprint of a sibling installation fails. Therefore we reduced the cases in which the blueprint is resolved to those where it is really needed. 

Fixes #

We tested the fix against the failing landscaper installation for the dev landscape and the problems were resolved.

**Release note**:

```bugfix operator
- root installation do not fail anymore if a sibling root installation fails to resolve its blueprint
```

### DIFF
--- a/pkg/landscaper/installations/imports/validation_test.go
+++ b/pkg/landscaper/installations/imports/validation_test.go
@@ -85,7 +85,7 @@ var _ = Describe("Validation", func() {
 		Expect(op.ResolveComponentDescriptors(ctx)).To(Succeed())
 
 		op.Context().Parent = inInstRoot
-		op.Context().Siblings = []installations.InstallationBaseInterface{inInstA}
+		op.Context().Siblings = []*installations.InstallationBase{inInstA.InstallationBase}
 		Expect(op.SetInstallationContext(ctx)).To(Succeed())
 
 		val := imports.NewValidator(op)
@@ -128,7 +128,7 @@ var _ = Describe("Validation", func() {
 		Expect(err).ToNot(HaveOccurred())
 
 		op.Context().Parent = inInstRoot
-		op.Context().Siblings = []installations.InstallationBaseInterface{inInstA}
+		op.Context().Siblings = []*installations.InstallationBase{inInstA.InstallationBase}
 		Expect(op.SetInstallationContext(ctx)).To(Succeed())
 
 		val := imports.NewValidator(op)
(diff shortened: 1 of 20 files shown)


pkg/landscaper/installations/installation.go
```diff
@@ -7,27 +7,35 @@ package installations
...
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
 	"github.com/gardener/landscaper/pkg/landscaper/blueprints"
 )
 
-// Installation is the internal representation of a installation
-type Installation struct {
-	Imports   map[string]interface{}
-	Info      *lsv1alpha1.Installation
-	Blueprint *blueprints.Blueprint
+type InstallationBaseInterface interface {
```
schrodit, line 16: not sure if we really need the interface.
I think just having the `InstallationBase` struct is enough.

schrodit (changes requested):

pkg/landscaper/installations/installation.go
achimweigel: We introduced the interface to be able to provide Installations and InstallationBases to methods which only need objects fulfilling the contract of InstallationBaseInterface. Of course we could go another way by removing the interface and changing the contract of those functions to InstallationBase, but then we must convert Installations first into InstallationBases when these are called.

achimweigel: we removed the interface

pkg/landscaper/installations/imports/validation_test.go
```diff
@@ -141,16 +141,16 @@ var _ = Describe("Validation", func() {
...
 		Expect(err).ToNot(HaveOccurred())
-		inInstA.Info.Status.Phase = lsv1alpha1.ComponentPhaseProgressing
-		Expect(fakeClient.Update(ctx, inInstA.Info)).To(Succeed())
+		inInstA.GetInfo().Status.Phase = lsv1alpha1.ComponentPhaseProgressing
+		Expect(fakeClient.Update(ctx, inInstA.GetInfo())).To(Succeed())
 
 		inInstB, err := installations.CreateInternalInstallation(ctx, op, fakeInstallations["test1/b"])
 		Expect(err).ToNot(HaveOccurred())
-		inInstB.Info.Status.Phase = lsv1alpha1.ComponentPhaseSucceeded
+		inInstB.GetInfo().Status.Phase = lsv1alpha1.ComponentPhaseSucceeded
```
schrodit, line 149: As we moved to the struct approach I think we do not need that getter indirection anymore.
So we could just leave most of the code as it is
```suggestion
		inInstB.Info.Status.Phase = lsv1alpha1.ComponentPhaseSucceeded
```

pkg/landscaper/installations/installation.go
```diff
@@ -107,6 +96,42 @@ func (i *Installation) IsImportingTarget(name string) bool {
...
 
 // MergeConditions updates or adds the given condition to the installation's condition.
-func (i *Installation) MergeConditions(conditions ...lsv1alpha1.Condition) {
+func (i *InstallationBase) MergeConditions(conditions ...lsv1alpha1.Condition) {
 	i.Info.Status.Conditions = lsv1alpha1helper.MergeConditions(i.Info.Status.Conditions, conditions...)
 }
+
+// Installation is the internal representation of a installation
+type Installation struct {
+	*InstallationBase
```
schrodit, line 105: why is the pointer needed?
I think a pointer here makes only sense if both structs could live on their own in parallel.
Or if the installation base could be nil but that could lead to more problems.

I propose that we just embed the struct.
```suggestion
	InstallationBase
```

schrodit (changes requested):

schrodit:
close in favor of https://github.com/gardener/landscaper/pull/197

PR #197, opened 2021-05-14
Not fetching the blueprint of installations if not needed

schrodit:

/area control-plane
/kind bug
/priority 3

PR is based on #194  with applied review comments.

A root installation fails if the resolution of the blueprint of a sibling installation fails. Therefore we reduced the cases in which the blueprint is resolved to those where it is really needed.

Fixes #

**Release note**:

```bugfix user
Installations do not fail anymore if a sibling installation fails to resolve its blueprint
```


schrodit:
Test issue is not related.
/lgtm

PR #200, opened 2021-05-17
Fix target annotation and label propagation

### LINKED ISSUE #199: Target Annotation and Label is not correctly synced

/area landscaper
/kind bug
/priority 3

**What happened**:

When creating a target with annotations or labels and the target is used in a subinstallation the annotations and labels are lost.
This is needed as deployer could rely on annotations or labels for their responsibility.

The annotations and labels are correctly synced.

schrodit:

/area landscaper
/kind bug
/priority 3

Fixes a bug where the targets annotations and labels are not correctly synced.

Also added a unit tests for the complete target sync.

Fixes https://github.com/gardener/landscaper/issues/199

**Release note**:

```bugfix user
Annotations and labels defined on targets are now correctly synced to an installations context.
```

### DIFF
--- a/pkg/landscaper/dataobjects/target.go
+++ b/pkg/landscaper/dataobjects/target.go
@@ -7,8 +7,11 @@ package dataobjects
 import (
 	"encoding/json"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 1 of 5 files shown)


Diaphteiros (approved):
/lgtm

PR #201, opened 2021-05-17
default crd management using k8s defaulting

schrodit:

/kind cleanup
/priority 3

Currently we rely on the helm chart defaulting for the crd management.
But with that local development is not considered and the helm does not contains all necessary defaults.

Therefore, the defaulting has been updated to use the standard k8s defaulting in the crd's.

Also the updating mechanism has been improved by now using patch calls.

Fixes #

**Release note**:

```feature operator
The defaulting mechanism for the CRD management has been updated to use the default k8s defaulting.
Note: the golang configuration struct has slightly changed.
```


Diaphteiros (approved):
/lgtm

PR #203, opened 2021-05-19
Fix force-reconcile deadlock in execution controller

schrodit:

/area operations
/kind bug
/priority 3

This PR fixes a bug were the execution controller and the deploy items controller compete on the `reconcile` annotation.
The execution constantly sets the reconcile annotation on the deploy item if the execution has the `force-reconcile` annotation. This reconcile only stops when all deployitems items are successfully reconciled.

The deployer now constantly tries to remove the reconcile annotation which ends in reconcile errors so that the deploy item can never be set to `Succeeded`.

```
Operation cannot be fulfilled on deployitems.landscaper.gardener.cloud \"my-deployitem\": the object has been modified; please apply your changes to the latest version and try again
```

This PR fixes that deadlock by removing the force annotation as soon as all deployitems are successfully updated.
Similar can still happen if the deployitems can never be updated. There should be a follow up tracked in this issue https://github.com/gardener/landscaper/issues/204 .

**Release note**:

```bugfix operator
A bug has been fixed that caused the a deadlink the `force-reconcile` logic of the execution controller and the deployers.
```

### DIFF
--- a/pkg/landscaper/execution/reconcile.go
+++ b/pkg/landscaper/execution/reconcile.go
@@ -106,14 +106,6 @@ func (o *Operation) Reconcile(ctx context.Context) error {
 
 	o.exec.Status.ObservedGeneration = o.exec.Generation
 
-	if !lsv1alpha1helper.IsCompletedExecutionPhase(phase) {
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


achimweigel (approved):
lgtm

PR #209, opened 2021-05-28
Improve testsetup by using the new testenv tool

### LINKED ISSUE #208: Use new EnvTest Binaries Setup Tool

/area dev-productivity
/kind enhancement
/priority 4

**Description**:

Currently we install the test tools (kubebuilder apiserver + etcd) directly from the kubebuilder bucket using curl.
The controller-runtime has introduced a new tool manage these binaries also to have multiple versions available.
So lets switch our install requirements script to that new tool.

https://github.com/kubernetes-sigs/controller-runtime/pull/1488

schrodit:

/area dev-productivity
/priority 3

Replaces the curl based testenv setup with the new controller-runtime testenv tool.
This new tool should make it easier in the future to update to newer k8s versions or even test different versions.

Fixes https://github.com/gardener/landscaper/issues/208

**Release note**:

```other operator
NONE
```

### DIFF
--- a/hack/setup-testenv.sh
+++ b/hack/setup-testenv.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+K8S_VERSION="1.19.x"
+
+echo "> Setup Test Environment for K8s Version ${K8S_VERSION}"
+
+CURRENT_DIR=$(dirname $0)
+PROJECT_ROOT="${CURRENT_DIR}"/..
+export KUBEBUILDER_ASSETS=$(setup-envtest use -p path ${K8S_VERSION})
+mkdir -p ${PROJECT_ROOT}/tmp/test
+rm -f ${PROJECT_ROOT}/tmp/test/bin
+ln -s "${KUBEBUILDER_ASSETS}" ${PROJECT_ROOT}/tmp/test/bin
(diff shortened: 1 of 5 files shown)


hack/setup-testenv.sh
```diff
@@ -0,0 +1,18 @@
...
+
+K8S_VERSION="1.19.x"
+
+echo "> Setup Test Environment for K8s Version ${K8S_VERSION}"
+
+CURRENT_DIR=$(dirname $0)
+PROJECT_ROOT="${CURRENT_DIR}"/..
+export KUBEBUILDER_ASSETS=$(setup-envtest use -p path ${K8S_VERSION})
+mkdir -p ${PROJECT_ROOT}/tmp/test
+rm -f ${PROJECT_ROOT}/tmp/test/bin
```
achimweigel, line 17:
```suggestion
rm -rf ${PROJECT_ROOT}/tmp/test/bin
```

test/utils/envtest/environment.go
```diff
@@ -44,15 +44,18 @@ func New(projectRoot string) (*Environment, error) {
...
 	testBinPath := filepath.Join(projectRoot, "tmp", "test", "bin")
-	if err := os.Setenv("TEST_ASSET_KUBE_APISERVER", filepath.Join(testBinPath, "kube-apiserver")); err != nil {
-		return nil, err
-	}
-	if err := os.Setenv("TEST_ASSET_ETCD", filepath.Join(testBinPath, "etcd")); err != nil {
-		return nil, err
-	}
-	if err := os.Setenv("TEST_ASSET_KUBECTL", filepath.Join(testBinPath, "kubectl")); err != nil {
-		return nil, err
+	if _, err := os.Stat(testBinPath); err == nil {
```
achimweigel, line 47: The case `os.Stat(testBinPath) != nil` is not handled

achimweigel (changes requested):

hack/setup-testenv.sh
schrodit: the `/tmp/test/bin` is just a symbolic link so I do not need to recursively remove it?

test/utils/envtest/environment.go
schrodit: that is by intention because if the testbindir is not set we rely on the kubebuilder defaults.
I added a comment to explain that behavior

achimweigel (approved):

PR #212, opened 2021-05-28
improve force reconcile deadlock in the execution controller

schrodit:

/area ops-productivity
/kind bug
/priority 3

Fixes a deadlock in the execution that can occur when the execution is in a Final state.
The force reconcile annotation now also resets the phase even if the generation stays the same.

**Release note**:

```bugfix operator
Fixes the `force-reconcile` handling in the execution controller that could lead to a reconcile deadlock.
```

### DIFF
--- a/pkg/landscaper/controllers/execution/controller.go
+++ b/pkg/landscaper/controllers/execution/controller.go
@@ -114,6 +114,19 @@ func HandleAnnotationsAndGeneration(ctx context.Context, log logr.Logger, c clie
 		}
 		log.V(7).Info("successfully updated metadata")
 	}
+
+	// also reset the phase when the force reconcile annotation is present.
+	// Otherwise we would never bbe able to leave a final phase
+	if lsv1alpha1helper.HasOperation(exec.ObjectMeta, lsv1alpha1.ForceReconcileOperation) {
+		if lsv1alpha1helper.IsCompletedExecutionPhase(exec.Status.Phase) {
+			exec.Status.Phase = lsv1alpha1.ExecutionPhaseInit
+			log.V(7).Info("updating status")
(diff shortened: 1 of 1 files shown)


pkg/landscaper/controllers/execution/controller.go
```diff
@@ -114,6 +114,19 @@ func HandleAnnotationsAndGeneration(ctx context.Context, log logr.Logger, c clie
 		}
 		log.V(7).Info("successfully updated metadata")
 	}
+
+	// also reset the phase when the force reconcile annotation is present.
+	// Otherwise we would never bbe able to leave a final phase
+	if lsv1alpha1helper.HasOperation(exec.ObjectMeta, lsv1alpha1.ForceReconcileOperation) {
```
robertgraeff, line 120: This works. But why does this method handle the force reconcile different than the normal reconcile? Wouldn't it be simpler to include the force reconcile case in the condition in line 94?
```
if hasReconcileAnnotation || hasForceReconcileAnnotation || exec.Status.ObservedGeneration != exec.Generation {
```

schrodit: I also tried that first but there is a problem with your suggested approach.

The problem is the different annotation handling.
The reconcile annotation is directly removed in this function as soon as the status is set.
Which means that event with additional reconciles we do not reset the phase.

The force-reconcile annotation is handled differently as it may stay for multiple reconcile loops.
Which means that this function could be called multiple times within one force-reconcile loop which would lead to always update/reset the status.
Thats why we only want to update if the status is in a final state. To get it out of the final state but not update progressing deploy items

robertgraeff (approved):
lgtm

PR #213, opened 2021-05-28
Fix missing defaulting for Landscaper Controller config

MrBatschner:

/area quality
/kind bug
/priority 3

The configuration object was not properly initialized with default values which caused the landscaper controller to crash if it was started without a configuration file (due to uninitialized NIL pointers). This PR makes sure the configuration object gets initialized with default values if not configuration file is supplied. Builds upon PR #201.

**Release note**:

```bugfix operator
Proper initialization of configuration with default values, fixes crash if landscaper-controller is started without configuration file.
```


schrodit (approved):
/lgtm

MrBatschner:
Forgot to add the context to the call to `opts.Complete(ctx)` in `app_test.go`.
/ok-to-merge

PR #214, opened 2021-05-28
improve deploy item abort timeout handling

schrodit:

/area TODO
/kind robustness
/priority 3

Improves the abort timeout handling that can cause a infinite reconcile loop between the deployer and the landscaper deployitem controller if a aborted deployitem is reconciled.
The issue is that the deployitem controller does not remove the abort timeout annotation which causes the deploy item controller to read the abort failed state whereas the deployer in parallel tries to reconcile the item.

This PR removes that behavior so that the abort annotations are removed as soon as the deploy item controller sets the deploy item into a failed state.

@Diaphteiros do we have other annotation that are also not cleaned up correctly? Or should the deployers also clean up these other annotations.

**Release note**:

```bugfix operator
Improved the abort handling of the deploy item controller to not cause an infinite reconcile while competing with a deployer
```

### DIFF
--- a/apis/core/v1alpha1/helper/helpers.go
+++ b/apis/core/v1alpha1/helper/helpers.go
@@ -51,9 +51,21 @@ func SetTimestampAnnotationNow(obj *metav1.ObjectMeta, ta TimestampAnnotation) {
 	metav1.SetMetaDataAnnotation(obj, string(ta), time.Now().Format(time.RFC3339))
 }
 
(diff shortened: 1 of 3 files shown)


Diaphteiros (approved):
/lgtm

If I remember correctly, the only other annotation that I introduced regarding the timeouts is the timestamp for the deployer pickup, which is to be removed by the corresponding deployer and not by the landscaper itself, so it should be fine.

PR #215, opened 2021-05-31
improve handling of multiple concurrent managers by using errgroups

### LINKED ISSUE #192: Improve error handling with multiple manager

/area robustness
/kind enhancement
/priority 5

Improve the error handling in the deployer lib if multiple managers are started (`StartManagers`).
Currently the whole controller is terminated if an error occurs in one of the managers.
But such errors should be handled more gracefully.

schrodit:

/area robustness
/kind enhancement
/priority 3

Improves the error handling for multiple concurrently running managers within different go routines.

Fixes https://github.com/gardener/landscaper/issues/192

**Release note**:

```other operator
NONE
```

### DIFF
--- a/cmd/landscaper-agent/app/app.go
+++ b/cmd/landscaper-agent/app/app.go
@@ -11,6 +11,7 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/gardener/landscaper/pkg/agent"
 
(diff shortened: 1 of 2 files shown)


cmd/landscaper-agent/app/app.go
```diff
@@ -49,20 +50,26 @@ func (o *options) run(ctx context.Context) error {
...
+		o.log.Info("Cache of host cluster successfully synced")
 	}
-	return nil
+	eg.Go(func() error {
+		if err := o.LsMgr.Start(ctx); err != nil {
+			return fmt.Errorf("error while running landscaper manager: %w", err)
+		}
+		return nil
+	})
+	return eg.Wait()
```
achimweigel, line 74: It is not clear for me what you want to achieve with this construct, perhaps lack of knowledge. According to the documentation  `eg.Wait()` waits until all function calls returned. Could it not happen that e.g. `o.HostMgr.Start` fails but `o.LsMgr.Start` is still running resulting in a system with only one of the two manager working.

schrodit: no this cannot happen because the context of `eg, ctx := errgroup.WithContext(ctx)` will get automatically canceled if one of the functions started with `eg.Go` return an error.

See the errgroups docs:
```
// WithContext returns a new Group and an associated Context derived from ctx.
//
// The derived Context is canceled the first time a function passed to Go
// returns a non-nil error or the first time Wait returns, whichever occurs
// first.
```

achimweigel (approved):
/lgtm

PR #218, opened 2021-06-01
Add targetlist support

Diaphteiros:

/area operations
/kind enhancement
/priority 3

This PR adds support for TargetList imports: Instead of importing a single target, blueprints can now also declare imports of multiple targets. In the installation, the targets can be selected either by listing their names or by using a label selector.

**Release note**:

```feature user
It is now possible to import lists of targets. Further information can be found in the documentation for [installations](docs/usage/Installations.md#targetlist-imports) and [blueprints](docs/usage/Blueprints.md#targetlists).
```
```bugfix user
Fixed a bug where the validation would not detect duplicate import names in installations across import types (e.g. a data and a target import with the same name).
```


apis/core/v1alpha1/types_installation.go
```diff
@@ -313,21 +341,38 @@ type SecretLabelSelectorRef struct {
 type ImportStatusType string
 
 const (
-	DataImportStatusType   ImportStatusType = "dataobject"
+	// DataImportStatusType is an ImportStatusType for data objects
+	DataImportStatusType ImportStatusType = "dataobject"
+	// TargetImportStatusType is an ImportStatusType for targets
 	TargetImportStatusType ImportStatusType = "target"
+	// TargetListImportStatusType is an ImportStatusType for target lists
+	TargetListImportStatusType ImportStatusType = "targetlist"
```
schrodit, line 349:
```suggestion
	TargetListImportStatusType ImportStatusType = "targetList"
```

docs/usage/Blueprints.md
```diff
@@ -179,6 +179,19 @@ imports:
...
+
+For some scenarios, it might be helpful to import an unknown number of targets. This can be done via a targetlist import. To declare a targetlist import, simply set `targetListType` instead of `targetType` in the import definition:
+
+```yaml
+imports:
+- name: my-import
+  targetListType: <string> # name of the target type for all targets in the list
+```
+
+Either `targetType` or `targetListType` has to be set, not both. Also, all targets of a targetlist need to have the same type, mixing is not possible.
```
schrodit, line 192:
```suggestion
Either `targetType` or `targetListType` has to be set, not both. Also, all targets of a targetlist *MUST* have the same type, mixing is not possible.
```

pkg/landscaper/dataobjects/targetlist.go
```diff
@@ -0,0 +1,85 @@
...
+// 		return fmt.Errorf("unable to apply target data: list size mismatch: internal %d - argument %d", len(t.Raw), len(raw))
+// 	}
+// 	for i := range t.Raw {
+// 		raw[i].Name = lsv1alpha1helper.GenerateDataObjectName(t.Metadata.Context, t.Metadata.Key)
+// 		raw[i].Namespace = t.Metadata.Namespace
+// 		raw[i].Spec = t.Raw[i].Spec
+// 		SetMetadataFromObject(raw[i], t.Metadata)
+// 	}
+// 	return nil
+// }
```
schrodit, line 85: can be removed if not needed

pkg/landscaper/dataobjects/targetlist.go
```diff
@@ -0,0 +1,85 @@
...
+package dataobjects
+
+import (
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
+)
+
+// Target is the internal representation of a target.
+type TargetList struct {
+	Targets    []*Target
+	FieldValue *lsv1alpha1.FieldValueDefinition
```
schrodit, line 14: the field value is not needed if all values are tracked by the referenced targets?

pkg/landscaper/installations/helper.go
```diff
@@ -174,6 +174,45 @@ func GetTargetImport(ctx context.Context, op lsoperation.Interface, contextName
...
+	owners := make([]*v1.OwnerReference, len(targetNames))
+	for i, targetName := range targetNames {
+		// get deploy item from current context
+		raw := &lsv1alpha1.Target{}
+		targetName = helper.GenerateDataObjectName(contextName, targetName)
+		if err := op.Client().Get(ctx, kubernetes.ObjectKey(targetName, inst.Info.Namespace), raw); err != nil {
+			return nil, nil, err
+		}
+		targets[i] = *raw
+		owners[i] = kubernetes.GetOwner(raw.ObjectMeta)
```
schrodit, line 189: maybe it makes sense to have the owner as part of the dataobject?
WDYT?

schrodit (changes requested):
The imported targets hsould also be available in the templating.
So the targets should be accessible as import .e.g `{{ .imports.my-import[0] }}`.
otherwise the imported targets could never be used.

pkg/landscaper/installations/helper.go
Diaphteiros: I like this idea, will have a look at it.

apis/core/v1alpha1/helper/dataobjects.go
```diff
@@ -42,6 +42,10 @@ func GenerateDataObjectName(context string, name string) string {
 	return base32.NewEncoding(Base32EncodeStdLowerCase).WithPadding(base32.NoPadding).EncodeToString(h.Sum(nil))
 }
 
+func GenerateDataObjectNameWithIndex(context string, name string, index int) string {
```
schrodit, line 45: docstring is missing

pkg/landscaper/dataobjects/targetlist.go
```diff
@@ -0,0 +1,107 @@
...
+	"encoding/json"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
+	lsv1alpha1helper "github.com/gardener/landscaper/apis/core/v1alpha1/helper"
+	kutil "github.com/gardener/landscaper/pkg/utils/kubernetes"
+)
+
+// Target is the internal representation of a target.
```
schrodit, line 17: wrong docstring

pkg/landscaper/dataobjects/targetlist.go
```diff
@@ -0,0 +1,107 @@
...
+type TargetList struct {
+	Targets []*Target
+}
+
+// New creates a new internal targetlist.
```
schrodit, line 22: wrongdocstring

pkg/landscaper/dataobjects/targetlist.go
```diff
@@ -0,0 +1,107 @@
...
+func NewTargetList() *TargetList {
+	return &TargetList{
+		Targets: []*Target{},
+	}
+}
+
+// New creates a new internal targetlist.
```
schrodit, line 29: wrong docstring

schrodit (changes requested):

Diaphteiros:
@schrodit I added a bugfix to this PR: In the installation validation, we only checked for duplicate import names in the same import category, meaning we did not catch it if there was a target import with the same name as a data import. Since we merge all imports - independently of type - into one map in [this method](https://github.com/gardener/landscaper/blob/v0.10.0/pkg/landscaper/installations/imports/constructor.go#L132), we should make sure that the import names are unique across all import categories/types. I adapted the validation accordingly with https://github.com/gardener/landscaper/pull/218/commits/cedf33d5c03e35bccaccd5ec1e3e6aab3d069a7a

apis/core/validation/blueprint.go
```diff
@@ -400,18 +423,21 @@ func ValidateInstallationTemplate(fldPath *field.Path, template *core.Installati
...
+	allErrs = append(allErrs, tmpErrs...)
+	tmpErrs, _ = ValidateInstallationTargetImports(imports.Targets, fldPath.Child("targets"), importNames)
+	allErrs = append(allErrs, tmpErrs...)
 
 	return allErrs
 }
 
 // ValidateInstallationTemplateDataImports validates the data imports of an InstallationTemplate
-func ValidateInstallationTemplateDataImports(imports []core.DataImport, fldPath *field.Path) field.ErrorList {
+func ValidateInstallationTemplateDataImports(imports []core.DataImport, fldPath *field.Path, importNames map[string]bool) (field.ErrorList, map[string]bool) {
```
schrodit, line 438: can you use a `sets.StringSet`. It is essentially the same implementation but seems more readable to me

apis/core/validation/installation_test.go
```diff
@@ -233,10 +233,28 @@ var _ = Describe("Installation", func() {
...
+							"t1",
+							"t2",
+							"t3",
+						},
+					},
+					{
+						Name:    "fb",
+						Targets: []string{},
+					},
+				},
```
schrodit, line 257: the test only covers duplicated target names but not a data import and target with the same name?
Or do I miss something here?

pkg/landscaper/dataobjects/target.go
```diff
@@ -96,7 +106,7 @@ func (t Target) Build() (*lsv1alpha1.Target, error) {
 	return newTarget, nil
 }
 
-// Apply applies data and metadata to a existing target.
+// Apply applies data and metadata to a existing target (except owner references).
```
schrodit, line 109:
```suggestion
// Apply applies data and metadata to an existing target (except owner references).
```

pkg/landscaper/installations/helper.go
```diff
@@ -157,22 +157,76 @@ func GetDataImport(ctx context.Context, kubeClient client.Client, contextName st
...
+	for k, v := range selector {
+		selectorWithContext[k] = v
+	}
+	if len(contextName) != 0 {
+		// top-level targets probably don't have an empty context set, so only add the selector if there actually is a context
+		selectorWithContext[lsv1alpha1.DataObjectContextLabel] = contextName
+	}
+	if err := kubeClient.List(ctx, targets, client.MatchingLabels(selectorWithContext), client.InNamespace(inst.Info.Namespace)); err != nil {
+		return nil, err
+	}
```
schrodit, line 209: you can also do the have a selector that gets all target in a namespace that have not set that label.

If we would use such a selector we would not have to loop through them ourself

pkg/landscaper/installations/imports/constructor.go
```diff
@@ -115,10 +119,41 @@ func (c *Constructor) constructImports(importList lsv1alpha1.ImportDefinitionLis
...
+				return nil, installations.NewErrorf(installations.SchemaValidationFailed, err, "%s: imported target does not match the expected target template schema", defPath.String())
 			}
 			if def.TargetType != targetType {
-				return nil, installations.NewErrorf(installations.SchemaValidationFailed, nil, "%s: exported target type is %s but expected %s", defPath.String(), targetType, def.TargetType)
+				return nil, installations.NewErrorf(installations.SchemaValidationFailed, nil, "%s: imported target type is %s but expected %s", defPath.String(), targetType, def.TargetType)
+			}
+			continue
+		case lsv1alpha1.ImportTypeTargetList:
+			if val, ok := templatedDataMappings[def.Name]; ok {
+				imports[def.Name] = val
```
schrodit, line 130: this should not be allowed for targets

schrodit (changes requested):

apis/core/validation/installation_test.go
Diaphteiros: The lines you have linked belong to the test that "should pass if imports are valid". It does actually not contain any duplicates (I needed to change some import names, since it did include cross-type duplicates before and wouldn't pass anymore after my changes to the validation).
The test that checks for duplicates starts in line 246.

pkg/landscaper/installations/helper.go
Diaphteiros: Ah, I didn't find out how to configure a "negative" selector and thought this was not possible. Will check again.

pkg/landscaper/installations/helper.go
```diff
@@ -157,22 +159,89 @@ func GetDataImport(ctx context.Context, kubeClient client.Client, contextName st
...
+	targetList, err := dataobjects.NewFromTargetList(targets.Items)
+	if err != nil {
+		return nil, err
 	}
-	return target, owner, nil
+	return targetList, nil
 }
 
+// GeReferenceFromComponentDescriptorDefinition tries to extract a component descriptor reference from a given component descriptor definition
```
schrodit, line 244: double doc string?

pkg/landscaper/installations/imports/constructor.go
```diff
@@ -64,7 +68,7 @@ func (c *Constructor) Construct(ctx context.Context, inst *installations.Install
 }
 
 // constructImports is an auxiliary function that can be called in a recursive manner to traverse the tree of conditional imports
-func (c *Constructor) constructImports(importList lsv1alpha1.ImportDefinitionList, importedDataObjects map[string]*dataobjects.DataObject, importedTargets map[string]*dataobjects.Target, templatedDataMappings map[string]interface{}, fldPath *field.Path) (map[string]interface{}, error) {
+func (c *Constructor) constructImports(importList lsv1alpha1.ImportDefinitionList, importedDataObjects map[string]*dataobjects.DataObject, importedTargets map[string]*dataobjects.Target, importedTargetLists map[string]*dataobjects.TargetList, templatedDataMappings map[string]interface{}, fldPath *field.Path) (map[string]interface{}, error) {
```
schrodit, line 71: can you restructure the imports list to be more readable on smaller screens:)
```suggestion
func (c *Constructor) constructImports(
  importList lsv1alpha1.ImportDefinitionList, 
  importedDataObjects map[string]*dataobjects.DataObject, 
  importedTargets map[string]*dataobjects.Target, 
  importedTargetLists map[string]*dataobjects.TargetList, 
  templatedDataMappings map[string]interface{}, 
  fldPath *field.Path) (map[string]interface{}, error) {
```

schrodit (changes requested):

pkg/landscaper/installations/imports/constructor.go
Diaphteiros: ah, always those people who don't use ultrawide monitors ;-)

schrodit (approved):
/lgtm

PR #221, opened 2021-06-02
Fix helm chart

Diaphteiros:

/area operations
/kind bug
/priority 2

Fixes a templating error in the helm chart and a small typo in an error message.
The templating error causes the rendered landscaper-config secret to contain an invalid configuration if the `landscaper.deployerManagement` field is specified in the values file, which prevents the landscaper from starting.

**Release note**:

```other operator
NONE
```


schrodit (approved):
/lgtm

PR #224, opened 2021-06-10
Typed import/export definitions

Diaphteiros:

/area quality
/kind api-change
/priority 2

Currently, data and target import definitions in a blueprint are distinguished by whether the `schema` or `targetType` field is specified. However, we plan on adding a few more types and the implicit distinction gets uglier with every new one, therefore we decided to add a `type` field which explicitly states the type of the import.

For backwards compatibility, we cannot make the field mandatory, as this would break all existing blueprints. Thus, the `type` field is optional for now, but will likely become mandatory with the next big api-change.

**Release note**:

```breaking user
A `type` field has been added to import and export definitions in blueprints. To preserve backward compatibility, it is currently optional. However, this is expected to change in the future, therefore it is strongly recommended to treat it as mandatory and always set it. Have a look at the [Blueprint documentation](docs/usage/Blueprints.md) for further information.
```

### DIFF
--- a/apis/core/types_blueprint.go
+++ b/apis/core/types_blueprint.go
@@ -11,6 +11,14 @@ import (
 // BlueprintResourceType is the name of the blueprint resource defined in component descriptors.
 const BlueprintResourceType = "blueprint"
 
+type ImportType string
+type ExportType string
+
+const ImportTypeData = ImportType("data")
+const ImportTypeTarget = ImportType("target")
+const ExportTypeData = ExportType(ImportTypeData)
+const ExportTypeTarget = ExportType(ImportTypeTarget)
+
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
@@ -67,6 +75,11 @@ type ImportDefinitionList []ImportDefinition
 type ImportDefinition struct {
 	FieldValueDefinition `json:",inline"`
 
+	// Type specifies which kind of object is being imported.
+	// This field should be set and will likely be mandatory in future.
+	// +optional
+	Type ImportType `json:"type,omitempty"`
+
 	// Required specifies whether the import is required for the component to run.
 	// Defaults to true.
 	// +optional
@@ -89,6 +102,11 @@ type ExportDefinitionList []ExportDefinition
 // ExportDefinition defines a exported value
 type ExportDefinition struct {
 	FieldValueDefinition `json:",inline"`
+
+	// Type specifies which kind of object is being exported.
+	// This field should be set and will likely be mandatory in future.
+	// +optional
+	Type ExportType `json:"type,omitempty"`
 }
 
 // FieldValueDefinition defines a im- or exported field.
--- a/apis/core/types_blueprint.go
+++ b/apis/core/types_blueprint.go
@@ -11,12 +11,22 @@ import (
 // BlueprintResourceType is the name of the blueprint resource defined in component descriptors.
 const BlueprintResourceType = "blueprint"
 
+// ImportType is a string alias
 type ImportType string
+
+// ExportType is a string alias
 type ExportType string
 
+// ImportTypeData is the import type for data imports
 const ImportTypeData = ImportType("data")
+
+// ImportTypeTarget is the import type for target imports
 const ImportTypeTarget = ImportType("target")
+
+// ExportTypeData is the export type for data exports
 const ExportTypeData = ExportType(ImportTypeData)
+
+// ExportTypeTarget is the export type for target exports
 const ExportTypeTarget = ExportType(ImportTypeTarget)
 
 // +genclient
--- a/apis/core/v1alpha1/defaults.go
+++ b/apis/core/v1alpha1/defaults.go
@@ -38,6 +38,10 @@ func SetDefaults_DefinitionImport(imports *ImportDefinitionList) {
 			// type is already set
 			continue
 		}
+		if imp.Schema != nil && len(imp.TargetType) != 0 {
+			// definition is invalid
+			continue
+		}
 		if imp.Schema != nil {
 			imp.Type = ImportTypeData
 		} else if len(imp.TargetType) != 0 {
@@ -55,6 +59,10 @@ func SetDefaults_DefinitionExport(exports *ExportDefinitionList) {
 		exp := &(*exports)[i]
 		if len(exp.Type) != 0 {
 			// type is already set
+			continue
+		}
+		if exp.Schema != nil && len(exp.TargetType) != 0 {
+			// definition is invalid
 			continue
 		}
 		if exp.Schema != nil {
--- a/apis/core/v1alpha1/defaults.go
+++ b/apis/core/v1alpha1/defaults.go
@@ -19,15 +19,49 @@ func SetDefaults_Blueprint(obj *Blueprint) {
 		obj.JSONSchemaVersion = "https://json-schema.org/draft/2019-09/schema"
 	}
 
-	for i := range obj.Imports {
-		SetDefaults_DefinitionImport(&(obj.Imports[i]))
-	}
+	SetDefaults_DefinitionImport(&obj.Imports)
+	SetDefaults_DefinitionExport(&obj.Exports)
 }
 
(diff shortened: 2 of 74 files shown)


docs/usage/Blueprints.md
```diff
@@ -137,12 +141,15 @@ The interface of a blueprint can be described using imports and exports.<br>
 The import definitions describe the data that needed by the blueprint.<br>
 The export definitions describe the data that is produced by the blueprint.<br>
 
-Imports and Exports always consists of a name (_MUST_ be unique for the blueprint) and can be generally described as _Target_-Im/Exports or _Data_-Im/Exports.
+Imports and Exports always consists of a name (_MUST_ be unique for the blueprint) and can be generally described as _Target_-Im/Exports or _Data_-Im/Exports. The `type` field specifies which one it is.
```
achimweigel, line 144:
```suggestion
Imports and Exports always consist of a name (_MUST_ be unique for the blueprint) and can be generally described as _Target_-Im/Exports or _Data_-Im/Exports. The `type` field specifies which one it is.
```

achimweigel (commented):
/lgtm only one little typo

pkg/landscaper/blueprints/blueprint.go
```diff
@@ -30,9 +30,46 @@ func New(blueprint *lsv1alpha1.Blueprint, content vfs.FileSystem) *Blueprint {
 		Info: blueprint,
 		Fs:   content,
 	}
+
+	// migration: set type field in import/export definitions if not set
+	// can be removed when the type field in import/export definitions is mandatory
+	migrateType(&b.Info.Imports)
```
schrodit, line 36: mhm not sure if we need that migration here.
I thought that we add the during the defaulting of blueprint (https://github.com/gardener/landscaper/blob/master/apis/core/v1alpha1/defaults.go#L17).

If we do it that we we would always have the type set which means we could always use the new code in the validation and in the handling. And would not require duplicated code or maintain old code path.

schrodit (changes requested):

pkg/landscaper/blueprints/blueprint.go
Diaphteiros: Good point, I didn't find the correct place to add the defaulting, but that one looks good, will move it there.

apis/core/v1alpha1/defaults.go
```diff
@@ -19,15 +19,49 @@ func SetDefaults_Blueprint(obj *Blueprint) {
...
+		SetDefaults_DefinitionImport(&imp.ConditionalImports)
+		if len(imp.Type) != 0 {
+			// type is already set
+			continue
+		}
+		if imp.Schema != nil {
+			imp.Type = ImportTypeData
+		} else if len(imp.TargetType) != 0 {
+			imp.Type = ImportTypeTarget
+		}
```
schrodit, line 45: here we should also handle the case that both types are defined.
maybe set a invalid type if both are defined.
Otherwise we would never run into the else part of the validation.

apis/core/v1alpha1/defaults.go
```diff
@@ -19,15 +19,49 @@ func SetDefaults_Blueprint(obj *Blueprint) {
...
+	if exports == nil {
+		return
+	}
+	for i := 0; i < len(*exports); i++ {
+		exp := &(*exports)[i]
+		if len(exp.Type) != 0 {
+			// type is already set
+			continue
+		}
+		if exp.Schema != nil {
```
schrodit, line 60: same issue as described above

apis/core/validation/blueprint.go
```diff
@@ -62,6 +73,31 @@ func validateBlueprintImportDefinitions(fldPath *field.Path, imports []core.Impo
 			importNames.Insert(importDef.Name)
 		}
 
+		if len(importDef.Type) != 0 {
+			// type is specified, use new validation
+			switch importDef.Type {
+			case core.ImportTypeData:
+				if importDef.Schema == nil {
```
schrodit, line 80: currently we ignore other fields.
Maybe we should check if other fields are defined that do not belong to the current type.

schrodit, line 80: currently we ignore other fields.
Maybe we should check if other fields are defined that do not belong to the current type.

pkg/landscaper/installations/exports/constructor.go
```diff
@@ -88,18 +88,21 @@ func (c *Constructor) Construct(ctx context.Context) ([]*dataobjects.DataObject,
 			continue
 		}
 
-		if def.Schema != nil {
+		switch def.Type {
```
schrodit, line 91: these cases now look much cleaner

schrodit (changes requested):

apis/core/validation/blueprint.go
Diaphteiros: I added a helper function that generically checks whether all config fields which are required for the current `Type` are set and that no other config fields are set.
The method itself doesn't look that nice (using the `reflect` package), but this way extending the import/export definitions will be much nicer and easier than if we would need to add exponentially more cases to ensure that no 'wrong' config field is set.

pkg/landscaper/installations/exports/constructor.go
Diaphteiros: Yes, I agree :-)

apis/core/types_blueprint.go
```diff
@@ -11,6 +11,14 @@ import (
 // BlueprintResourceType is the name of the blueprint resource defined in component descriptors.
 const BlueprintResourceType = "blueprint"
 
+type ImportType string
+type ExportType string
+
+const ImportTypeData = ImportType("data")
+const ImportTypeTarget = ImportType("target")
+const ExportTypeData = ExportType(ImportTypeData)
+const ExportTypeTarget = ExportType(ImportTypeTarget)
```
schrodit, line 20: can you add docstrings

apis/core/validation/blueprint.go
```diff
@@ -18,6 +22,47 @@ import (
...
+		}
+	}
+	for _, v := range exportTypesWithExpectedConfig {
+		for _, e := range v {
+			res[e] = true
+		}
+	}
+	return res
+}
+
```
schrodit, line 65: these are helper functions can we move them to the bottom or in a separate file?

apis/core/validation/blueprint.go
```diff
@@ -18,6 +22,47 @@ import (
...
+func keys(m map[string][]string) []string {
+	res := make([]string, len(m))
+	for k := range m {
+		res = append(res, string(k))
+	}
+	return res
+}
+
+func relevantConfigFields() map[string]bool {
```
schrodit, line 51: why is this function needed?
It returns always the same value?

schrodit (changes requested):

apis/core/validation/blueprint.go
Diaphteiros: Yes, it returns the union of the values of `importTypesWithExpectedConfig` and `exportTypesWithExpectedConfig`. We need this union to find out which config fields exist and thus need to be checked. For example: if the `type` is `target`, we expect `TargetType` to be set while `Schema` should not be set. We cannot simply verify that all other fields are not set, because some need to be set in any case, such as `Name` and `Type`.
Instead of using a positive list here, we could also turn it around and exclude `Name` and `Type` and then assume that all other fields are config and should not be set unless they are listed in the current value of `importTypesWithExpectedConfig`.

Do you get what I mean? WDYT?

Diaphteiros: Turning it the other way around will lead to the check function becoming uglier though, because then we will have to iterate over all fields of the import/export definition instead of iterating over a fixed set of fields.

schrodit: hm ok.
maybe it makes sense to expose that as a variable and only calculate that once during startup in the`init`func.
And can you add a docstring to describe the generates map

Diaphteiros: I just added a docstring, will have a look at computing the list only once during init.

Diaphteiros:
@schrodit Please have a look again.

schrodit (approved):
/lgtm

PR #134, opened 2021-02-26
Add resources readiness checks for Helm/Manifest deployments

MrBatschner:
**How to categorize this PR?**

/area quality
/kind enhancement
/priority normal

**What this PR does / why we need it**:

This commit adds Readiness checks for workload resource types based
on the resource status provided by Kubernetes.

It fixes a bug for the helm and manifests deployers that occurs when
trying to update resources with immutable field that are set on the
server side like ClusterIP or the ResourceVersion.

It fixes a bug for the helm and manifests deployers that
could occur during the deletion of the deploy items if the managed
resources in status have never been updated due to errors in the
Kubernetes manifests.

For the helm deployer, it removes the decoding of the status
when deleting the files because it is already done during the
creation of the new Helm struct.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

99,9% of the work in this PR is from @lcavajani and his PR #111 which will be closed in favour of this one.

This is the first part of #21 to keep the PR smaller.

**Release note**:

```
Add resources readiness checks for helm/manifest
```

PR #225, opened 2021-06-11
Custom readiness Checks

### LINKED ISSUE #21: Helm Deployer and Manifest Deployer Health checks

Currently the helm and manifest deployer just creates and updates the configured kubernetes resources.
It does validate if the deployment was successfull nor does it wait until the deployed resources are healthy.

It should be possible for developers to specify readiness/health checks so that the deploy items waits until the application is ready and possibly fail when the application does not get ready.

*Implementation Proposal*:

Health/readiness checks could be automatically added for known built-in resources like `Deployments`, `StatefulSets` or `Deamonsets`. (Could be hard to monitor if hpa/vpa is used)

A specific syntax is needed to describe also health and readiness checks for custom resources or "unknown" resources.
Maybe also for custom health checks on known resources.

**Proposed Configuration**:

```yaml
kind: ProviderConfiguration

healthChecks:
- resources:
    apiVersion: group/version
    kind: "" # should be single is name/namespace is used or the list object if the label selector is used
    name: ""
    namespace: ""
    labelSelector: {} # use default kubernetes label selectors
  timeout: 10s # duration
  checks:
  - jsonPath: "" # jsonpath in the object
    requirement: "notin (Progressing)" # kubernetes requirements could be used; see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors

readinessChecks: [] # same as healthChecks
```

In addition it should be also possible to specify specific error reporting fields.
So that failures reported by controllers can also be reported back to the deploy item (there also a generic solution to look for events for that objects are possible).

/kind enhancement
/area monitoring
/size m

MrBatschner:

/area deployers
/area robustness
/kind enhancement
/priority 3

PR #134 introduced readiness checks to the Helm and Manifest deployer which check core Kubernetes resources for their readiness. They are - however - not capable of performing checks on custom fields or on CRs.

This PR introduces custom readiness checks (still called `HealthChecks` for compatibility) which offer fine granular configuration of which resources need to be checked and which of their fields should be matched against given values.

Resources can be identified...
- by their GVK, namespace and name
- by their GVK and a set of labels

Fields, identified by their JSON path...
- can be matched to exactly one value
- can be matched to any value in a list
- can be matched to no value in a list
- must exist (no matter with what value)
- must not exist

Multiple custom healthchecks can be specified for different resource conditions, they can be individually enabled and disabled.

Fixes #21 

This PR is huge because using `k8s.io/client-go/util/jsonpath` in the apis-submodule pulled in lots of dependencies.

**Release note**:

```other operator
introduces custom readiness checks which offer fine granular configuration of which fields of which resources should be checked against which values
```


apis/deployer/healthchecks/doc.go
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors.
```
schrodit, line 1: think we should move them to `apis/deployer/utils/healthchecks` so that we do not mix them with other main deployer configurations

apis/deployer/healthchecks/doc.go
```diff
@@ -0,0 +1,11 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package core is the internal version of the API.
+// +k8s:deepcopy-gen=package
+// +k8s:openapi-gen=true
+// +k8s:defaulter-gen=TypeMeta
+
+// +groupName=healthchecks.deployer.landscaper.gardener.cloud
```
schrodit, line 10: I think this is not needed as we do not have specific typed objects but these are rather helper structs

apis/deployer/healthchecks/types_healthchecks.go
```diff
@@ -0,0 +1,63 @@
...
+	// Timeout is the time to wait before giving up on a resource to be healthy.
+	// Defaults to 180s.
+	// +optional
+	Timeout *lsv1alpha1.Duration `json:"timeout,omitempty"`
+	// CustomHealthChecks is a set of custom health check configurations
+	// +optional
+	CustomHealthChecks []CustomHealthCheckConfiguration `json:"customHealthChecks,omitempty"`
+}
+
+type CustomHealthCheckConfiguration struct {
```
schrodit, line 28: docstring is missing

apis/deployer/healthchecks/types_healthchecks.go
```diff
@@ -0,0 +1,63 @@
...
+	// +optional
+	Resource *lsv1alpha1.TypedObjectReference `json:"resourceSelector,omitempty"`
+	// Labels are the labels used to identify multiple resources that can be identified by a unique set of labels
+	// +optional
+	LabelSelector *LabelSelectorSpec `json:"labelSelector,omitempty"`
+	// Requirements is the actual health-check which compares an object's property to a value
+	Requirements []RequirementSpec `json:"requirements"`
+}
+
+type LabelSelectorSpec struct {
```
schrodit, line 46: docstring missing

apis/deployer/healthchecks/validation/validation.go
```diff
@@ -0,0 +1,121 @@
...
+
+// ValidateHealthCheckConfiguration validates a healthcheck configuration
+func ValidateHealthCheckConfiguration(fldPath *field.Path, config *healthchecks.HealthChecksConfiguration) field.ErrorList {
+	var allErrs field.ErrorList
+
+	// if we have a custom healthcheck configuration, the default should be disabled
+	disableDefault := config.DisableDefault
+	customHealthChecks := config.CustomHealthChecks
+	if !disableDefault && customHealthChecks != nil && len(customHealthChecks) > 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("customHealthChecks"), disableDefault, "custom health checks require disabling the default health check"))
```
schrodit, line 25: why do we require the default to be disabled?

I think its valid to say that I still want default healthchecks for deployment and still have custom healthchecks for crds.
Maybe we should also make the default healthchecks configurable per resource? WDYT?

apis/deployer/healthchecks/types_healthchecks.go
```diff
@@ -0,0 +1,63 @@
...
+	// DisableDefault allows to disable the default health checks.
+	// +optional
+	DisableDefault bool `json:"disableDefault,omitempty"`
+	// Timeout is the time to wait before giving up on a resource to be healthy.
+	// Defaults to 180s.
+	// +optional
+	Timeout *lsv1alpha1.Duration `json:"timeout,omitempty"`
+	// CustomHealthChecks is a set of custom health check configurations
+	// +optional
+	CustomHealthChecks []CustomHealthCheckConfiguration `json:"customHealthChecks,omitempty"`
```
schrodit, line 25:
```suggestion
	CustomHealthChecks []CustomHealthCheckConfiguration `json:"custom,omitempty"`
```
I think `healthchecks` might be duplicated as we are already in the healtcheck configuration struct

apis/deployer/healthchecks/validation/validation.go
```diff
@@ -0,0 +1,121 @@
...
+		selection.NotEquals,
+		selection.In,
+		selection.NotIn,
+	}
+
+	for _, op := range allowedOperators {
+		if operator == op {
+			return nil
+		}
+	}
```
schrodit, line 119: nit maybe you can use a switch statement. then the allocation and loop is not required
```suggestion
	var allowedOperators = []selection.Operator{
		selection.DoesNotExist,
		selection.Exists,
		selection.Equals,
		selection.DoubleEquals,
		selection.NotEquals,
		selection.In,
		selection.NotIn,
	}

	switch op {
	case selection.DoesNotExist, ...:
	  return nil
	 default:
	   return err
	 }
```

apis/deployer/healthchecks/validation/validation.go
```diff
@@ -0,0 +1,121 @@
...
+	return errors.New("not an allowed operator")
```
schrodit, line 120: the err value is never used so we could return a boolean

apis/deployer/healthchecks/validation/validation_test.go
```diff
@@ -0,0 +1,254 @@
...
+			DisableDefault:     true,
+			Timeout:            &lsv1alpha1.Duration{Duration: 180 * time.Second},
+			CustomHealthChecks: []healthchecks.CustomHealthCheckConfiguration{cc},
+		}
+
+	})
+
+	AfterEach(func() {
+		defer ctx.Done()
+	})
```
schrodit, line 86: not needed. (should be also removed in the other tests)

pkg/deployer/helm/ensure.go
```diff
@@ -119,10 +119,35 @@ func (h *Helm) ApplyFiles(ctx context.Context, files map[string]string, exports
...
+			Log:              h.log,
+			Timeout:          h.ProviderConfiguration.HealthChecks.Timeout,
+			ManagedResources: h.ProviderStatus.ManagedResources,
+		}
+		err := defaultHealthCheck.CheckResourcesHealth()
 		if err != nil {
 			return err
 		}
+	} else if h.ProviderConfiguration.HealthChecks.CustomHealthChecks != nil {
+		//TODO: thread this
```
schrodit, line 135: with thread you mean parallelize?

pkg/deployer/lib/healthcheck/customhealthcheck.go
```diff
@@ -0,0 +1,245 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
+	lsv1alpha1helper "github.com/gardener/landscaper/apis/core/v1alpha1/helper"
+	"github.com/gardener/landscaper/apis/deployer/healthchecks"
+	"github.com/gardener/landscaper/pkg/utils"
+	kutil "github.com/gardener/landscaper/pkg/utils/kubernetes"
+)
+
+type CustomHealthCheck struct {
```
schrodit, line 31: missing docstring

pkg/deployer/lib/healthcheck/customhealthcheck.go
```diff
@@ -0,0 +1,245 @@
...
+	Context          context.Context
+	Client           client.Client
+	Log              logr.Logger
+	CurrentOp        string
+	Timeout          *lsv1alpha1.Duration
+	ManagedResources []lsv1alpha1.TypedObjectReference
+	Configuration    healthchecks.CustomHealthCheckConfiguration
+}
+
+func (c *CustomHealthCheck) CheckResourcesHealth() error {
```
schrodit, line 41: missing docstring

pkg/deployer/lib/healthcheck/customhealthcheck.go
```diff
@@ -0,0 +1,245 @@
...
+
+		if found {
+			results = append(results, obj)
+		}
+	}
+
+	return results, nil
+}
+
+func GetFieldsByJSONPath(obj map[string]interface{}, fieldPath string) ([][]reflect.Value, error) {
```
schrodit, line 203: missing docstring

schrodit (changes requested):
first look I need some more time for the implementation review

apis/deployer/healthchecks/doc.go
MrBatschner: Done

apis/deployer/healthchecks/doc.go
MrBatschner: Done

apis/deployer/healthchecks/types_healthchecks.go
MrBatschner: Fixed

apis/deployer/healthchecks/types_healthchecks.go
MrBatschner: Fixed

pkg/deployer/helm/ensure.go
MrBatschner: Yeah, I meant to put the custom health checks into go() functions so that they can run in parallel instead of sequentially. However, the custom health check calls `WaitForObjectsHealthy()` which already is parallelized (it calls `IsObjectHealthy() inside a go func(){} block) and therefore, I believe it is safe to just remove this comment. Therefore, I do not see any benefit in kicking off the custom healthchecks in parallel and and I will just remove the comment.

pkg/deployer/lib/healthcheck/customhealthcheck.go
MrBatschner: Fixed.

pkg/deployer/lib/healthcheck/customhealthcheck.go
MrBatschner: Fixed.

pkg/deployer/lib/healthcheck/customhealthcheck.go
MrBatschner: Fixed.

pkg/deployer/helm/ensure.go
MrBatschner: Comment removed.

apis/deployer/healthchecks/validation/validation_test.go
MrBatschner: Fixed.

apis/deployer/healthchecks/validation/validation.go
MrBatschner: Yes, totally makes sense. I remove this piece of code from the validation and changed the `ensure.go` of both, Manifest and Helm deployer so that default and custom healthchecks can run together.

apis/deployer/utils/healthchecks/types_healthchecks.go
```diff
@@ -0,0 +1,66 @@
...
+	APIVersion string `json:"apiVersion"`
+	// Kind is the Kind of the object to be selected by labels
+	Kind string `json:"kind"`
+	// Labels are the labels used to identify multiple resources of the given kind
+	Labels map[string]string `json:"matchLabels"`
+}
+
+// RequirementSpec contains the requirements an object must meet to pass the custom health check
+type RequirementSpec struct {
+	JsonPath string             `json:"jsonPath"`
```
schrodit, line 59: can you add a docstring that explains in more detail the format of the jsonpath.

schrodit: Currently it is required to use the syntax with surrounding `{}`. 
This is needed by the underlying k8s jsonpath lib. 
Maybe it makes sense to reuse this lib function: https://github.com/gardener/landscaper/blob/master/pkg/landscaper/dataobjects/jsonpath/jsonpath.go#L19 that
automatically injects the needed syntax.
So then it would be possible to describe a jsonpath as `.status.constraint` instead of `{.status.constraint}`:
Which may seem more intuitive. WDYT?

pkg/deployer/lib/healthcheck/defaulthealthcheck.go
```diff
@@ -2,144 +2,59 @@
...
-
-	if len(allErrs) > 0 {
-		return apimacherrors.NewAggregate(allErrs)
-	}
-	if len(notHealthyErrs) > 0 {
-		return apimacherrors.NewAggregate(notHealthyErrs)
-	}
-
-	return nil
+type DefaultHealthCheck struct {
```
schrodit, line 24: docstring missing

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -248,6 +248,14 @@ func ObjectFromTypedObjectReference(ref *v1alpha1.TypedObjectReference) *unstruc
 	}
 }
 
+// ObjectKeyFromTypedObjectReference returns the object key (NamespacedName) from a TypedObjectReference
+func ObjectKeyFromTypedObjectReference(obj *v1alpha1.TypedObjectReference) *types.NamespacedName {
```
schrodit, line 252: isn't this function already available on the v1alpha1.ObjectReference?

schrodit (changes requested):
Implementation looks good. just some minor things.

Can you add some docs about the health check? (I thought that you have already written some but I'm unable to find it in the PR?)

apis/deployer/utils/healthchecks/types_healthchecks.go
MrBatschner: Good idea but that function does not have `allowMissingKeys` enabled which I make use of. I just took the few lines to add the `{}` and potentially missing leading `.` to the JSON path to make the usage much simpler.

pkg/utils/kubernetes/kubernetes.go
MrBatschner: I am not even using it anywhere... Removed.

MrBatschner:
> Implementation looks good. just some minor things.
> 
> Can you add some docs about the health check? (I thought that you have already written some but I'm unable to find it in the PR?)

Added an additional file under docs that outlines the healthchecks.

apis/deployer/utils/healthchecks/validation/validation_test.go
```diff
@@ -0,0 +1,249 @@
...
+		hc healthchecks.HealthChecksConfiguration
+		cc healthchecks.CustomHealthCheckConfiguration
+		rc healthchecks.RequirementSpec
+	)
+
+	BeforeEach(func() {
+		fld = field.NewPath("healthcheck")
+
+		rc = healthchecks.RequirementSpec{
+			JsonPath: "{ .foo.bar }",
```
schrodit, line 42:
```suggestion
			JsonPath: ".foo.bar",
```

docs/deployer/healthchecks.md
```diff
@@ -0,0 +1,112 @@
...
+## Custom Health Checks
+
+Custom health checks can be used to match custom fields of selected resources to given values.
+
+Multiple custom health checks can be defined for DeployItem as multiple resources to be checked each require a specifially tailored custom health check.
+
+A custom health check must contain a name and a selector that selects the resource to be checked. This selector can either be:
+
+- a set of `resourceSelector` each of which matches just one resource at a time, identified by its `apiVersion`, `kind`, `namespace` and `name`
+- a `labelSelector` that matches multiple resources of the same `apiVersion` and `kind`, identified by a set of labels they need to have
```
schrodit, line 99: I think we should mention that only resources that are deployed by the deployitem are matched.
Otherwise people might think that with a label selector all resources in the cluster are checked.

schrodit (changes requested):
Code looks good.

just a last question regarding the api.
We currently name the check healthcheck.
Not sure if we should name it readinessCheck and then be able to add liveness checks when we have real reconcile.
With that the transition would be easier and we would align with the K8s naming in pods.

docs/deployer/healthchecks.md
MrBatschner: That piece of information was a bit hidden, it is now explicitly mentioned (twice!).

MrBatschner:
> Code looks good.
> 
> just a last question regarding the api.
> We currently name the check healthcheck.
> Not sure if we should name it readinessCheck and then be able to add liveness checks when we have real reconcile.
> With that the transition would be easier and we would align with the K8s naming in pods.

Yes, we already talked about that. My take is that I would like to do this in a separate PR. Consistently changing the names of the health check throughout the code requires touching lots of files and this PR is already way too big.

schrodit (approved):
/lgtm

PR #229, opened 2021-06-16
Improve landscaper blueprint handling

### LINKED ISSUE #187: Improve Blueprint handling in the landscaper

/area robustness
/kind technical-debt
/priority 2

Currently blueprints are only cached using the default oci caching.
But with that we only cache the tar version. 
Everytime a installation is reconciled that tar is extracted to a temporary directory.

We currently also get the blueprints for all sibling/parent installations which is not needed for reconciliation.
So we should only fetch the blueprint if really necessary.

**Why is this needed**:
We should have a dedicated blueprint cache that directly caches the extracted blueprints.

schrodit:

/area quality
/kind enhancement
/priority 3

A blueprint cache has been introduced in the Landscaper.
Previously every fetch of a blueprint has resulted in a new temporary directory which led to a lot of network traffic and disk size.
The cache now caches blueprints and shares them across reconciles.
A Blueprint is identified using the component name, component version, resource name and resource version.
The cache can also be disabled which means that a blueprint is always fetched but it is still shared. This should only be used for development as blueprints are not immutable anymore.

Fixes https://github.com/gardener/landscaper/issues/187

**Release note**:

```feature operator
A blueprints cache has been introduced in the Landscaper which improves the handling of Blueprints internally.
The Cache size and garbage collection can be configured in the Landscaper configuration.
```


pkg/landscaper/blueprints/resolve.go
```diff
@@ -80,60 +73,75 @@ func Resolve(ctx context.Context, resolver ctf.ComponentResolver, cdRef *lsv1alp
...
 		return nil, err
 	}
 
-	var data bytes.Buffer
-	// todo: do not store the blueprint in memory and rather directly decode the bytestream to the filesystem.
-	blobInfo, err := blobResolver.Resolve(ctx, resource, &data)
+	if blueprint, err := GetStore().Get(ctx, cd, resource); err == nil {
+		return blueprint, nil
+	}
+	// blueprint was not cached so we need to fetch the blueprint blob and store it in the cache
```
achimweigel, line 96: Don't we need to check the error returned by `GetStore().Get(ctx, cd, resource)` like

`if error != NotFoundError {
  return nil, err
}`

pkg/landscaper/blueprints/store.go
```diff
@@ -0,0 +1,303 @@
...
+
+var storeSingleton *Store
+var mux sync.Mutex
+
+// GetStore returns the currently active store.
+func GetStore() *Store {
+	mux.Lock()
+	defer mux.Unlock()
+	return storeSingleton
+}
```
achimweigel, line 46: What do you want to achieve with the mutex in this method? You will still conflict with SetStore because you get a pointer to a Store which might be closed just after you fetched it.  But perhaps I misunderstood this part.

pkg/landscaper/blueprints/store.go
```diff
@@ -0,0 +1,303 @@
...
+		if err := s.fs.RemoveAll(bpPath); err != nil {
+			s.log.Error(err, "unable to cleanup directory")
+			return nil, errors.New("unable to cleanup store")
+		}
+	}
+
+	if err := s.fs.Mkdir(bpPath, os.ModePerm); err != nil && !os.IsExist(err) {
+		return nil, fmt.Errorf("unable to create bluprint directory: %w", err)
+	}
+	s.mux.Lock()
```
achimweigel, line 176: Why does the mutex not include the s.fs cleanup at the beginning of the method?

achimweigel (changes requested):

pkg/landscaper/blueprints/store.go
schrodit: ah yes good point. The cleanup and `create dir` should be included in the mutex

pkg/landscaper/blueprints/store.go
schrodit: Yes you're right I could get a pointer to a closed store.
Maybe I also need to check that and return errors when the store is already closed.

As for the mutex: I first wanted to have a real singleton like 
```
func GetStore() *Store {
  if store == nil {
    store = createNewStore()
  }
  return store
}
```
In that case a mutex was necessary to no create multiple stores at once.
But I decided against it as the store needs more error handling like temp dirs defaulting etc..
Thats why a default store is not initialized in the init function.

But after that you're right that the new mutex does not really give any more safety.
I will remove it and will return proper errors for closed stores

pkg/landscaper/blueprints/resolve.go
schrodit: I explicitly did not handled that error as I'm was unsure what to do about such errors.
In this case errors could happen due to corrupted blueprints or temporary filesystem failures.
In both I wanted the application to give a recovery chance by re-downloading the blueprint and restoring it in the store.

Both cases could have negative side effect:
- either we are in a infinite reconcile loop due to a corrupted disk/corrupted blueprint with no chance to recover
- or we are downloading the blueprint to often (but still would get the error handled in line 133 when the blueprint is tried to be stored)

WDYT how should we handle such issues?

achimweigel (approved):
/lgtm 

only one small typo

PR #234, opened 2021-06-25
Improve execution reporting

### LINKED ISSUE #179: Propagate Deploy Item Phase up to Execution/Installation

/area robustness
/kind bug
/priority 3

With #143 and #174, the landscaper will set the phase of deploy items to `Failed` under certain circumstances. Currently, this is not propagated to the executions and installations, so if a deploy item fails due to a timeout, the execution and installation it belongs to will still show up as `Progressing` instead of `Failed`. 

The desired behaviour would be that the execution and installation, which created the deploy item, are also set to `Failed` if the deploy item fails.

schrodit:

/area operations
/kind bug
/priority 3

Fixes a bug that a failed state is not propagated when there is no deployer that reconciles a deploy item.

Fixes https://github.com/gardener/landscaper/issues/179

This PR depends on https://github.com/gardener/landscaper/pull/233 as also differnt error states are now reported though events

**Release note**:

```bugfix user
A bug has been fixed, that caused a Installation to hang in `Progressing` state although the corresponding deploy item is already failed.
The state is now correctly propagated and displayed to the user.
```

### DIFF
--- a/apis/core/v1alpha1/constants.go
+++ b/apis/core/v1alpha1/constants.go
@@ -67,3 +67,11 @@ const (
 	// LandscaperMetricsNamespaceName describes the prometheus metrics namespace for the landscaper component
 	LandscaperMetricsNamespaceName = "ociclient"
 )
+
+// DeployItem care controller constants
+const (
+	PickupTimeoutReason      = "PickupTimeout"    // for error messages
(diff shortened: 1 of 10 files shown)


Diaphteiros (approved):
/lgtm

PR #237, opened 2021-07-02
fix blueprint store directory name length

schrodit:

/area robustness
/kind bug
/priority 3

Fixes a bug where the generated blueprint id for the store exceeded max allowed name for a file/directory.
This PR fixes this by always using a hex encoded `sha256` hash of the id.

**Release note**:

```bugfix user
A bug has been fixed that caused the blueprint store to panic when the generated blueprint id exceeded the max allowed characters for a directory name.
```

### DIFF
--- a/pkg/landscaper/blueprints/store.go
+++ b/pkg/landscaper/blueprints/store.go
@@ -6,7 +6,8 @@ package blueprints
 
 import (
 	"context"
-	"encoding/base32"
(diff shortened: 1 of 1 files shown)


MrBatschner (approved):
/lgtm

Diaphteiros (approved):
/lgtm

PR #240, opened 2021-07-06
Fix CRD generation

schrodit:

/kind bug
/priority 3

Fixes a bug where the CRD resources could not be generated anymore.
We do now not rely on the controller-tools anymore and use the already generated k8s openapi to generate the resources.
Additional information for the crd like `group`, `version`, `primary storage` is directly configured as go struct in the api definitions.

Fixes #

**Release note**:

```other operator
NONE
```


Diaphteiros (approved):
/lgtm

PR #242, opened 2021-07-08
improve deployer logging and print config on startup

schrodit:

/area deployers
/kind quality
/priority 3

Improves the logging (debugging) for the deployers regarding configuration and target selectors

**Release note**:

```feature operator
The logging of deployers has been improved so that the configuration and additional target selector information can be logged with the according verbosity level.
```

### DIFF
--- a/pkg/deployer/lib/controller.go
+++ b/pkg/deployer/lib/controller.go
@@ -154,28 +154,16 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	var target *lsv1alpha1.Target
-	if deployItem.Spec.Target != nil {
-		target = &lsv1alpha1.Target{}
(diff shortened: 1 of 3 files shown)


Diaphteiros (approved):
/lgtm

pkg/deployer/lib/controller.go
```diff
@@ -154,8 +154,15 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	var target *lsv1alpha1.Target
+	target, responsible, err := c.checkTargetResponsibility(ctx, logger, deployItem)
```
robertgraeff, line 157: The check of the target is now done twice. First in the new method `checkTargetResponsibility`, and then a second time after the method was called.

robertgraeff (changes requested):

pkg/deployer/lib/controller.go
schrodit: good catch

robertgraeff (approved):

PR #245, opened 2021-07-09
Add tests

schrodit:

/kind test
/priority 3

added tests for different execution and installation failures

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/landscaper/controllers/execution/reconcile_test.go
+++ b/pkg/landscaper/controllers/execution/reconcile_test.go
@@ -1,0 +1,85 @@
+// SPDX-FileCopyrightText: 2020 SAP SE or an SAP affiliate company and Gardener contributors.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package execution_test
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/gardener/landscaper/pkg/api"
+	"github.com/gardener/landscaper/pkg/landscaper/controllers/execution"
+	kutil "github.com/gardener/landscaper/pkg/utils/kubernetes"
+	testutils "github.com/gardener/landscaper/test/utils"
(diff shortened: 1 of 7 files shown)


pkg/landscaper/controllers/installations/reconcile_delete_test.go
```diff
@@ -134,4 +149,35 @@ var _ = Describe("Delete", func() {
...
+		_ = testutils.ShouldNotReconcile(ctx, ctrl, testutils.RequestFromObject(inst)) // returns a still waiting error
+
+		exec := &lsv1alpha1.Execution{}
+		exec.Name = "root"
+		exec.Namespace = state.Namespace
+		testutils.ExpectNoError(testenv.Client.Get(ctx, kutil.ObjectKeyFromObject(exec), exec))
+		Expect(exec.DeletionTimestamp).ToNot(BeNil())
+
+		testutils.ExpectNoError(testenv.Client.Get(ctx, kutil.ObjectKeyFromObject(inst), inst))
+		metav1.SetMetaDataAnnotation(&exec.ObjectMeta, lsv1alpha1.OperationAnnotation, string(lsv1alpha1.ForceReconcileOperation))
```
achimweigel, line 172: I'm a little bit confused. Why setting the annotation at the execution and not the installation if you want to test its propagation?

schrodit: thanks good catch.
Obviously that test succeeds.

When I fixed that test it then fails. So I will also fix the Landscaper regarding that issue

pkg/landscaper/controllers/execution/reconcile_test.go
```diff
@@ -0,0 +1,84 @@
...
+		di := &items.Items[0]
+		//set item to failed state
+		di.Status.Phase = lsv1alpha1.ExecutionPhaseFailed
+
+		// then reconcile the execution and delete it
+		testutils.ShouldReconcile(ctx, ctrl, testutils.RequestFromObject(exec))
+		testutils.ExpectNoError(testenv.Client.Delete(ctx, exec))
+		testutils.ShouldReconcile(ctx, ctrl, testutils.RequestFromObject(exec))
+
+		testutils.ShouldReconcile(ctx, ctrl, testutils.RequestFromObject(exec))
```
jschicktanz, line 78: You have the same call 2 lines above which I assume is for checking that the reconcile works after the Delete operation. Why check the same thing here again?

schrodit: we have to reconcile 2 times because on the first reconcile the deployitem is deleted and on the second reconcile the execution is deleted.

jschicktanz (approved):

robertgraeff (approved):

Issue #246, opened 2021-07-12
Deployer Target Owner Object


schrodit:
**How to categorize this issue?**

/area deployers
/kind enhancement
/priority 4

**Description**

Currently deployers identify their ownership for deployitem based on the target's labels and annotations.
This is could be hard to debug and is only configurable once per deployer.

To be able to also update the ownership during runtime and per namespace/target, a new possibility to configure the ownership during runtime with a user (operating) facing resource should be introduced.

The new object should be optional and should be added to the deployer lib as their can be a lot of reuse potential.

Subtasks:
- [ ] Add Generic `Owner` object crd. https://github.com/gardener/landscaper/pull/247
- [ ] Refactor CRD management as generic crd manager
- [ ] Implement owner reconcile in deployer lib

achimweigel:
@In-Ko closed because outdated and/or copied into internal project

PR #247, opened 2021-07-12
add deployer owner resource

schrodit:

/area deployers
/kind enhancement
/priority 3

First part of https://github.com/gardener/landscaper/issues/246 .
Implements the shared custom owner crd.

Also enhance the crd generation so that crd's can be written to different directories

**Release note**:

```other operator
NONE
```


pkg/deployer/lib/crdresources/deployer.landscaper.gardener.cloud_owners.yaml
```diff
@@ -0,0 +1,89 @@
...
+                      description: Name is the name of the kubernetes object.
+                      type: string
+                    namespace:
+                      description: Namespace is the namespace of kubernetes object.
+                      type: string
+                  required:
+                  - name
+                  type: object
+                type: array
+              type:
```
achimweigel, line 53: Just a general question. If the deployer ID defines a deployer, why do we need the type field?

achimweigel (commented):
If ownership objects could be modified doesn't this result in a quite complicated synchronisation problem if one deployer is just working on an item and another is taking over?

pkg/deployer/lib/crdresources/deployer.landscaper.gardener.cloud_owners.yaml
schrodit: you're right that the type field is not necessary but its more for readability.
So that human operators can better correlate.

schrodit:
> If ownership objects could be modified doesn't this result in a quite complicated synchronisation problem if one deployer is just working on an item and another is taking over?

Owner objects are not for synchronisation. Their main purpose is that the responsibility for targets could be modified during runtime.
Also these owner objects are not part of the Landscaper but part of the deployer. So e.g. they live in the fenced cluster when the deployer is deployed their.
So they are just local configuration.

With the current approach, using annotations or labels, we have the same issue. There could be multiple deployers reacting to the same deploy items.

apis/deployer/core/types.go
```diff
@@ -0,0 +1,50 @@
...
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// OwnerList contains a list of Owners.
+type OwnerList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Owner `json:"items"`
```
enrico-kaack-comp, line 19: The key 'items' contains Owners, seems unintuitive to me, is it intended?

enrico-kaack-comp (approved):
lgtm

apis/deployer/core/types.go
schrodit: yes this is required from the k8s api convention. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds

Issue #243, opened 2021-07-08
No deployment triggered by changed import values in a configmap or secret 


robertgraeff:
**How to categorize this issue?**

/area robustness
/kind bug
/priority 3

**What happened**:
Consider an Installation that references a configmap or secret in its `.imports.data` section. Suppose the Installation is succeeded. Then change the data in the configmap (resp. secret). This change does not trigger a new deployment.

**What you expected to happen**:
A change of import values should trigger a new deployment.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
The reason might be that a configmap / secret does not have a generation number - as opposed to a DataObject.

**Environment**:

- Landscaper version: v0.10.0
- Kubernetes version (use `kubectl version`):
- Others:

robertgraeff:
Also the annotation `landscaper.gardener.cloud/operation: reconcile` did not help to trigger a new deployment.

The annotation `landscaper.gardener.cloud/operation: force-reconcile` triggered to a new deployment.

PR #268, opened 2021-07-27
Fix data reconcile trigger for secrets and configmaps

schrodit:

/area operations
/kind bug
/priority 3

Fixes #https://github.com/gardener/landscaper/issues/243

**Release note**:

```bugfix user
A bug has been fixed that caused the Landscaper to not detect data changes in secrets or configmaps
```

### DIFF
--- a/pkg/landscaper/installations/operation.go
+++ b/pkg/landscaper/installations/operation.go
@@ -172,7 +172,7 @@ func (o *Operation) GetImportedDataObjects(ctx context.Context) (map[string]*dat
 
 		var (
 			sourceRef *lsv1alpha1.ObjectReference
-			configGen = strconv.Itoa(int(do.Raw.Generation))
+			configGen = do.Metadata.Hash
 			owner     = kutil.GetOwner(do.Raw.ObjectMeta)
 		)
 		if owner != nil && owner.Kind == "Installation" {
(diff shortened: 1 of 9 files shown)


pkg/landscaper/installations/operation.go
```diff
@@ -172,7 +172,7 @@ func (o *Operation) GetImportedDataObjects(ctx context.Context) (map[string]*dat
 
 		var (
 			sourceRef *lsv1alpha1.ObjectReference
-			configGen = strconv.Itoa(int(do.Raw.Generation))
+			configGen = do.Metadata.Hash
```
Diaphteiros, line 175: You changed this only for `GetImportedDataObjects`, but `GetImportedTargets` follows the same logic - shouldn't this be changed there too?

Diaphteiros (changes requested):
I think you missed one line, but I might be wrong there. Looks good otherwise.

pkg/landscaper/installations/operation.go
schrodit: no because for targets it is still valid to use the generation field.
Only secrets and configmaps do not work with it

Diaphteiros: ah, ok, then it should be fine.

Diaphteiros (approved):
/lgtm

jschicktanz (approved):

PR #269, opened 2021-07-27
Fix pod volume permissions to not run as root

### LINKED ISSUE #264: Container Deploy Item: Permission denied when writing to shared-volume (`STATE_PATH`, `EXPORTS_PATH`)

/area container-deployer
/kind bug
/priority 2

**What happened**:

When trying to write a (custom) Container Deploy Item according to [this guide](https://github.com/gardener/landscapercli/blob/master/docs/commands/container_deployer/add_container_di.md), the writing of data to the `STATE_PATH` & `EXPORTS_PATH` (shared-volume) doesn't work because of some `Permissions denied` problem.

For example `echo "test" > $STATE_PATH/test.txt` will result in `/data/ls/shared/state/test.txt: Permission denied`.

For my thoughts on what the root cause could be, please take a look at the "Anything else we need to know?" section.

It should be possible to write to the shared-volume (`STATE_PATH` & `EXPORTS_PATH`) within a custom Container Deploy Item.

Try to run the script from [this guide](https://github.com/gardener/landscapercli/blob/master/docs/commands/container_deployer/add_container_di.md#the-example-script-program), which writes something to the shared-volume.

Or reproduce it by running this pod:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: init-permissions
spec:
  automountServiceAccountToken: false
  containers:
  - image: europe-west3-docker.pkg.dev/sap-se-gcp-istio-dev/public/create-file:latest
    name: main
    command:
    - ./script.sh
    env:
    - name: DIR_PATH
      value: /data/ls/shared
    volumeMounts:
    - mountPath: /data/ls/shared
      name: shared-volume
    securityContext:
      runAsUser: 1001
      runAsGroup: 1001
  initContainers:
  - image: europe-west3-docker.pkg.dev/sap-se-gcp-istio-dev/public/mkdir:latest
    name: init
    command:
    - ./script.sh
    env:
    - name: DIR_PATH
      value: /data/ls/shared
    volumeMounts:
    - mountPath: /data/ls/shared
      name: shared-volume
  volumes:
  - emptyDir: {}
    name: shared-volume
```

The `mkdir` container just runs:
```bash
#!/bin/sh
set -euo pipefail
echo "create directory 'test' in shared volume"
mkdir $DIR_PATH/test
```

And the `create-file` container runs:
```bash
#!/bin/sh
set -euo pipefail
echo "create file in directory"
echo "test" > $DIR_PATH/test/values.txt
```

I think the problem (as reproduced in the "How reproduce it" section) is that the init container runs with `root` and creates the `/data/ls/shared/state/` and the `/data/ls/shared/exports/values` directory/file with the `root` user. But the container provided by the developer runs with user `1001`, which is totally okay because this arbitrary code shouldn't run as `root`. But the user `1001` can't access the directories created by the init container (`root` user). To fix this problem, the init container could also run as user `1001`.

**Environment**:

- Landscaper version: `v0.10.0`
- Kubernetes version (use `kubectl version`): Server: `v1.19.12` (Gardener Cluster)
- Others: For the simple permissions test I've used k8s version `v1.21.1` (Gardener Cluster) to exclude problems of older k8s versions.

schrodit:

/area container-deployer
/kind bug
/priority 3

Currently the init and wait containers are executed as root user which is an issue as soon as the main execution container is not as root as the export and satte folder cannot be used anymore.
Therefore the container deployer now runs all container as the same user and group.
This also disabled the ability to run a container as root.

Fixes https://github.com/gardener/landscaper/issues/264

**Release note**:

```bugfix developer
A bug has been fixed that caused the container that are executed by the container-deployer to be unable to write to the export or state directory.
```

### DIFF
--- a/docs/deployer/container.md
+++ b/docs/deployer/container.md
@@ -94,8 +94,13 @@ When the image with your program is executed, it gets access to particular infor
 - If *componentDescriptor* and *blueprint* is specified, a *content blob* can be accessed at the directory given by 
   the env var `CONTENT_PATH`. The *content blob* consists of all data stored in the blueprint, consisting of the blueprint 
   yaml file and all other files and folders you stored together with this.
+- The container is not executed as root user due to security reasons.
+  Instead, the following user/group is used (see the [Kubernetes Documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod) for details):
+  - **User**: 1000
+  - **Group**: 3000
+  - **FSGroup**: 2000
   
-Again you will find a more detailed explanation of these env variable
+Again you will find a more detailed explanation of these env variables
 [here](https://github.com/gardener/landscapercli/blob/master/docs/commands/container_deployer/add_container_di.md).
 
 ### Status
--- a/pkg/deployer/container/pod.go
+++ b/pkg/deployer/container/pod.go
@@ -294,6 +294,11 @@ func generatePod(opts PodOptions) (*corev1.Pod, error) {
 	pod.Spec.RestartPolicy = corev1.RestartPolicyNever
 	pod.Spec.TerminationGracePeriodSeconds = pointer.Int64Ptr(300)
 	pod.Spec.Volumes = volumes
+	pod.Spec.SecurityContext = &corev1.PodSecurityContext{
+		RunAsUser:  pointer.Int64(1000),
+		RunAsGroup: pointer.Int64(3000),
+		FSGroup:    pointer.Int64(2000),
+	}
 	pod.Spec.InitContainers = []corev1.Container{initContainer}
 	pod.Spec.Containers = []corev1.Container{mainContainer, waitContainer}
 	if len(opts.ImagePullSecret) != 0 {


achimweigel (approved):
/lgtm

Diaphteiros (approved):

robertgraeff (approved):

loewenstein:
> This also disabled the ability to run a container as root.

Hi @schrodit,

just wondering if there are valid use cases for running a deploy container as root?
If my mental model is correct, the container is executed in the target cluster. So, given a Landscaper user has a use case for running as root, the blast radius would be that Landscaper user's cluster, right?
This might be an acceptable risk for allowing to run as root.

Best regards
Jan

achimweigel:
Hi @loewenstein, as far as I know, the container is executed on the cluster were the container deployer runs which is usually not the target cluster.

loewenstein:
@achimweigel yeah, I realized that looking into the code after a colleague told me.

I am wondering now, why the deploy item spec has a `target` key. Sure, that spec is provider independent, but it is documented in the container deployer specific readme.

achimweigel:
Hi @loewenstein, the target contains the access info of e.g. the target cluster and is accessible by your container when it is executed.

schrodit:
> I am wondering now, why the deploy item spec has a target key. Sure, that spec is provider independent, but it is documented in the container deployer specific readme.

This is because the target is used to determine the responsibility of a deployer.
In case of the container deployer, this is used for example if an something should be installed ina fenced environment. Then there is one container deployer running in the public cloud and on ein the fenced environment but we can install our application in both.

loewenstein:
> This is because the target is used to determine the responsibility of a deployer.

So the same configuration that is used to define where a manifest deployer deploys to is used for a completely different concern here? That sounds like a good source of confusion.

schrodit:
> > This is because the target is used to determine the responsibility of a deployer.
> 
> So the same configuration that is used to define where a manifest deployer deploys to is used for a completely different concern here? That sounds like a good source of confusion.

I think this is not an issue with the model but with the container deployer (whcih is a really specific deployer that does not follow the intended flow of deployers).
When we look at other deployers like the helm deployer, a target is always a kubernetes cluster. 
In this case multiple helm deployers can be responsible for different clusters which are then determined based on the cluster. 

The container deployer is really generic so there is no specific target. Also the target doesn't make that much sense as custom code is executed that can do anything even with multiple targets.
But still in the case of a container deployer we could have multiple environments therefore we said that we use the same strategy as for other deployer and do not introduce another concept just because of that.

PR #270, opened 2021-07-27
Fix nil references

robertgraeff:

/area robustness
/kind bug
/priority 3

This PR fixes some small issues I encountered when I started the landscaper controller with a small config like this:

```yaml
landscaper:
  registryConfig: # contains optional oci secrets
    allowPlainHttpRegistries: true
    secrets: {}
  deployers:
  - container
  - helm
  - manifest
```

(For example the section `deployItemTimeouts` is missing in my config. I am not sure which assumptions we make about the config.)

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/cmd/landscaper-controller/app/options.go
+++ b/cmd/landscaper-controller/app/options.go
@@ -73,23 +73,23 @@ func (o *Options) parseConfigurationFile(ctx context.Context) (*config.Landscape
 	decoder := serializer.NewCodecFactory(api.ConfigScheme).UniversalDecoder()
 
 	configv1alpha1 := &v1alpha1.LandscaperConfiguration{}
+
+	if len(o.ConfigPath) != 0 {
+		data, err := ioutil.ReadFile(o.ConfigPath)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, _, err := decoder.Decode(data, nil, configv1alpha1); err != nil {
+			return nil, err
+		}
+	}
+
 	api.ConfigScheme.Default(configv1alpha1)
+
 	config := &config.LandscaperConfiguration{}
 	err := api.ConfigScheme.Convert(configv1alpha1, config, ctx)
 	if err != nil {
-		return nil, err
-	}
-
-	if len(o.ConfigPath) == 0 {
-		return config, nil
-	}
-
-	data, err := ioutil.ReadFile(o.ConfigPath)
-	if err != nil {
-		return nil, err
-	}
-
-	if _, _, err := decoder.Decode(data, nil, config); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


achimweigel (approved):
/lgtm

MrBatschner:
> I am not sure which assumptions we make about the config.

Missing config elements should have been defaulted to meaningful, non-nil values in https://github.com/gardener/landscaper/blob/master/apis/config/v1alpha1/defaults.go, it might be worth checking if this actually works.

pkg/landscaper/blueprints/store.go
```diff
@@ -120,7 +120,7 @@ func NewStore(log logr.Logger, baseFs vfs.FileSystem, config config.BlueprintSto
 		gcConfig: config.GarbageCollectionConfiguration,
 	}
 
-	if config.Size != "0" {
+	if config.Size != "0" && config.Size != "" {
```
schrodit, line 123: hm the size should be also correctly defaulted as @MrBatschner mentioned

schrodit (changes requested):

pkg/landscaper/blueprints/store.go
robertgraeff: I have merged the default config and the provided config. The other initialisations from my first commit are then no longer necessary.

cmd/landscaper-controller/app/options.go
```diff
@@ -71,27 +71,29 @@ func (o *Options) Complete(ctx context.Context) error {
...
 	}
+
 	data, err := ioutil.ReadFile(o.ConfigPath)
 	if err != nil {
 		return nil, err
 	}
 
-	cfg := &config.LandscaperConfiguration{}
-	if _, _, err := decoder.Decode(data, nil, cfg); err != nil {
+	if _, _, err := decoder.Decode(data, nil, config); err != nil {
```
schrodit, line 92: I think the changes will not be merged but overwritten?
you could do, similar to above, first decode as v1alpha, then default and then convert to the core config.

schrodit (changes requested):

cmd/landscaper-controller/app/options.go
robertgraeff: I have debugged this. The default config contains for example the ls-system namespace twice in the deployerManagement:
```yaml
deployerManagement:
  namespace: ls-system
  agent:
    namespace: ls-system
```
Then `Decode` is applied with a custom config which contains only the second namespace:
```yaml
deployerManagement:
  agent:
    namespace: lndscpr-int-test
```
The result is a mixture of both:
```yaml
deployerManagement:
  namespace: ls-system
  agent:
    namespace: lndscpr-int-test
```
So, the fields of the default config which are not overwritten by the custom config seem to survive.

schrodit: hmm ok they survive but I think this is more due to weird internal struct overwrite.
I think this could lead to unintended behavior in the future (especially for slices or maps).
Therefore I would prefer to keep it clean and explicitly run the defaulting code.

robertgraeff: I have now changed it how you suggested. That's the better way.

schrodit (approved):
/lgtm

Issue #239, opened 2021-07-06
Improve Reconcile


schrodit:
**How to categorize this issue?**

/kind enhancement
/priority 3

Currently installations, executions and deployitems do not follow the reconcile pattern as they are only executed when the `reconcile-annotation` was set.
We should change that and allow deployers to actively reconcile deploy items and therefore constantly enforcing their state.

This new behavior leads other issues in the current flow.
For example that the state of a deploy item could be changed from `succeeded` to `failed` or the other way around which need to be propagated to its installation and parent.
If then also export change we need to trigger dependent installations.

To avoid flapping we should also have a state between `succeeded` and `failed` that indicates and error that may be self-healed.

Tasks:
- [ ] Propagate state from `succeeded` or `failed` deploy items to the installation
- [ ] trigger/reconcile dependent installations
- [ ] constantly reconcile deploy items https://github.com/gardener/landscaper/issues/103
- [ ] let the deployers constantly reconcile
  - [ ] https://github.com/gardener/landscaper/issues/35 - we should make it configurable if the deployers should constantly reconcile. Maybe also with an annotation that blocks constant reconcile.
  - [ ] Add constant reconcile to other deployers like the container deployer. There we might need to set a reconcile interval that specifies how often the given container should be re-executed.

PR #272, opened 2021-07-28
Properly reconcile DeployItems from Executions

### LINKED ISSUE #103: Reconcile deploy items

/area quality
/kind enhancement
/priority normal

Currently manually deleted deployitems are only reconciled when the installation resource is reconciled (either by updating the spec or setting the concile label).
If deployitems are deleted, the landscaper should automatically recreate the deployitems.

If that happens, also the subsequent deployitems as well as the installation's export execution should be re-reconciled.

Maybe also subsequent installations should be retriggered.

**Why is this needed**:
To be more resilient against deletion and more convenient for issues on deployment items.

MrBatschner:

/area robustness
/kind enhancement
/priority 3

So far, the landscaper-controller did not fully follow the K8S pattern of reconciliation - it would only reconcile dependent resources if the parent got modified in any way.
This PR changes the reconciliation logic of Execution -> DeployItem so that any changes to DeployItems will trigger the reconciler and make sure that the desired state (as described by the Execution) will match the actual state (as represented by the DeployItems).

one of mutliple PRs to deal with #239
Fixes #103 

**Release note**:

```other operator
Improves reconciliation logic for Executions and their dependent DeployItems.
```

### DIFF
--- a/pkg/landscaper/controllers/execution/controller.go
+++ b/pkg/landscaper/controllers/execution/controller.go
@@ -64,6 +64,10 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, errHdl(ctx, err)
 	}
 
+	if err := HandleDeployItemGeneration(ctx, c.client, logger, exec); err != nil {
+		return reconcile.Result{}, errHdl(ctx, err)
+	}
+
 	if lsv1alpha1helper.IsCompletedExecutionPhase(exec.Status.Phase) {
 		logger.V(7).Info("execution item is already in completion phase")
 		return reconcile.Result{}, nil
@@ -133,6 +137,34 @@ func HandleAnnotationsAndGeneration(ctx context.Context, log logr.Logger, c clie
 				return err
 			}
 			log.V(7).Info("successfully updated status")
+		}
+	}
+	return nil
+}
+
+// HandleDeployItemGeneration checks the managed DeployItems owned by this execution and will force a reconcilcation
+// by setting the Executions Phase to "Init" if an DeployItem cannot be retrieved or if a DeployItems's generation does not
+// match the observed generation.
+func HandleDeployItemGeneration(ctx context.Context, client client.Client, log logr.Logger, exec *lsv1alpha1.Execution) error {
+	for _, managedDi := range exec.Status.DeployItemReferences {
+		var err error
+		di := &lsv1alpha1.DeployItem{}
+
+		if err = client.Get(ctx, managedDi.Reference.NamespacedName(), di); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+
+		if (err != nil && apierrors.IsNotFound(err)) || managedDi.Reference.ObservedGeneration != di.Generation {
+			log.V(7).Info("deploy item cannot be found or does not match last observed generation")
+
+			exec.Status.Phase = lsv1alpha1.ExecutionPhaseProgressing
+
+			log.V(7).Info("updating status")
+			if err := client.Status().Update(ctx, exec); err != nil {
+				return err
+			}
+
+			break
 		}
 	}
 	return nil
--- a/pkg/landscaper/execution/reconcile.go
+++ b/pkg/landscaper/execution/reconcile.go
@@ -43,60 +43,63 @@ func (o *Operation) Reconcile(ctx context.Context) error {
 
 	var phase lsv1alpha1.ExecutionPhase
 	for _, item := range executionItems {
-		if item.DeployItem != nil && !o.forceReconcile {
-			phase = lsv1alpha1helper.CombinedExecutionPhase(phase, item.DeployItem.Status.Phase)
-			deployItemStatusUpToDate := item.DeployItem.Status.ObservedGeneration == item.DeployItem.GetGeneration()
-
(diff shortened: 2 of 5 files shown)


pkg/landscaper/controllers/execution/controller.go
```diff
@@ -138,6 +142,34 @@ func HandleAnnotationsAndGeneration(ctx context.Context, log logr.Logger, c clie
...
+			log.V(7).Info("deploy item cannot be found or does not match last observed generation")
+
+			exec.Status.Phase = lsv1alpha1.ExecutionPhaseProgressing
+
+			log.V(7).Info("updating status")
+			if err := client.Status().Update(ctx, exec); err != nil {
+				return err
+			}
+
+			break
```
schrodit, line 167: why do you need a break here?

pkg/landscaper/controllers/execution/controller.go
```diff
@@ -64,6 +64,10 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, errHdl(ctx, err)
 	}
 
+	if err := HandleDeployItemGeneration(ctx, c.client, logger, exec); err != nil {
+		return reconcile.Result{}, errHdl(ctx, err)
+	}
+
 	if lsv1alpha1helper.IsCompletedExecutionPhase(exec.Status.Phase) {
 		logger.V(7).Info("execution item is already in completion phase")
```
schrodit, line 72: I think we should only check the deployitem generation if we are in a completed phase?
otherwise we are already in a reconciling phase and we would just waste the resources and requests for all deploy items. WDYT?
```suggestion

	if lsv1alpha1helper.IsCompletedExecutionPhase(exec.Status.Phase) {
	    if err := HandleDeployItemGeneration(ctx, c.client, logger, exec); err != nil {
		    return reconcile.Result{}, errHdl(ctx, err)
	    }
		logger.V(7).Info("execution item is already in completion phase")
```

pkg/landscaper/controllers/execution/controller.go
```diff
@@ -138,6 +142,34 @@ func HandleAnnotationsAndGeneration(ctx context.Context, log logr.Logger, c clie
...
+// by setting the Executions Phase to "Init" if an DeployItem cannot be retrieved or if a DeployItems's generation does not
+// match the observed generation.
+func HandleDeployItemGeneration(ctx context.Context, client client.Client, log logr.Logger, exec *lsv1alpha1.Execution) error {
+	for _, managedDi := range exec.Status.DeployItemReferences {
+		var err error
+		di := &lsv1alpha1.DeployItem{}
+
+		if err = client.Get(ctx, managedDi.Reference.NamespacedName(), di); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
```
schrodit, line 155:
```suggestion
		di := &lsv1alpha1.DeployItem{}
		err := client.Get(ctx, managedDi.Reference.NamespacedName(), di)
		if err != nil && !apierrors.IsNotFound(err) {
			return err
		}
```

pkg/landscaper/execution/reconcile.go
```diff
@@ -152,6 +155,7 @@ func (o *Operation) deployOrTrigger(ctx context.Context, item executionItem) err
 		item.DeployItem = &lsv1alpha1.DeployItem{}
 		item.DeployItem.GenerateName = fmt.Sprintf("%s-%s-", o.exec.Name, item.Info.Name)
 		item.DeployItem.Namespace = o.exec.Namespace
+		item.DeployItem.Generation = o.exec.Generation
```
schrodit, line 158: the generation of a resource is updated by the k8s api server and cannot be updated by us.
Also the deploy items generation does not need to have the same generation as its execution.
```suggestion
```

pkg/landscaper/execution/reconcile_test.go
```diff
@@ -167,6 +167,22 @@ var _ = Describe("Reconcile", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(exec.Status.Phase).To(Equal(lsv1alpha1.ExecutionPhaseFailed))
 		})
+
+		It("should set the status of a previously failed execution to progressing if a deployitem has a newer generation than the last observed generation", func() {
```
schrodit, line 171: hmm not sure if this should always be the case.
The generation is only updated when something in the spec changes.
But we should also be able to reconcile status changes of the same generation.
Or is this is out-of-scope for this PR?

schrodit (changes requested):

pkg/landscaper/controllers/execution/controller.go
```diff
@@ -138,6 +142,34 @@ func HandleAnnotationsAndGeneration(ctx context.Context, log logr.Logger, c clie
 	return nil
 }
 
+// HandleDeployItemGeneration checks the managed DeployItems owned by this execution and will force a reconcilcation
+// by setting the Executions Phase to "Init" if an DeployItem cannot be retrieved or if a DeployItems's generation does not
```
Diaphteiros, line 146: It says `Init` in the docstring, but the phase is actually set to `ExecutionPhaseProgressing`, which is `Progressing`

Diaphteiros (changes requested):

Diaphteiros:
Closed in favour of #290

Issue #261, opened 2021-07-22
Support `exportsFromManifests` in Kubernetes Manifest Deployer


Haegi:
**How to categorize this issue?**

/area manifest-deployer
/kind enhancement
/priority 4

**What would you like to be added**:

Support exporting from manifest with `exportsFromManifests` in the Kubernetes Manifest Deployer like the [Helm Deployer does](https://github.com/gardener/landscaper/blob/master/docs/deployer/helm.md#provider-configuration).

**Why is this needed**:

 It would be great if it would also be possible to export resources/fields from the manifest(s) created with the Kubernetes Manifest Deployer.

loewenstein:
For better understanding, `exportFromManifest` of the Helm deployer seems to work solely on the rendered manifests, right?
I.e. there is no way to get to e.g. the `status` field of a resource, let alone a resource that was not directly rendered as part of the Helm Chart but created by e.g. an operator.

I'd say `exportFromManifest` and a mechanism to export data from arbitrary resources in the target cluster would be super useful for all deployers targeting clusters.

Haegi:
Yes, the dynamic imports/exports is a different and much harder problem.
At the moment the imports/exports are computed at the deploytime. One problem when having dynamic import/exports (e.g. LoadBalancer IPs) would be that you have to reconcile all the resources, which depend on this import/export and you don't know if it is the final state. There are more problems but as this is not related to this issue, we can maybe talk about it offline or you open a separate issue (maybe already with some idea/proposal).

schrodit:
> For better understanding, exportFromManifest of the Helm deployer seems to work solely on the rendered manifests, right?
yes this currently the case. During the initial implementation is was easier to just read from rendered manifests as reading from the status involves some more options to consider. Like timeouts or how to distinguish a not reconciled item from a wrong jsonpath.

But with the implementation of the readiness checks https://github.com/gardener/landscaper/pull/225 most of the code should be already there. And the implementation should be easier.

PR #273, opened 2021-07-30
Common manifest handling

schrodit:

/area deployers
/kind cleanup
/priority 3

Previously the code for reconciling kubernetes manifests in the manifest deployer and the helm deployer are independent although 90% of the code are identical.

Therefore now common api types and common manifest deployer is used for both deployers.
So that the code handling for both deployers should be easier in the future

Part of https://github.com/gardener/landscaper/issues/261

I also added the first draft api definition for exports from kubernetes manifests

**Release note**:

```feature developer
The kubernetes manifest handling of the manifest and helm deployer are combined and both deployers do now use the same apitypes and manifest handling.
```


Diaphteiros (approved):
/lgtm

PR #279, opened 2021-08-10
Add webhooks server commands line flags "webhook-url" and "certificates-namespace"

reshnm:

/area usability
/kind enhancement
/priority 3

Currently the landscaper webhook server needs to target in the same cluster where the webhook service is created.
When the webhook server is targeting a different cluster, the service reference that is currently being used, does no longer work.
For this use case an URL can be specified in the ValidatingWebhook configuration (see [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#url)), which references an _external_ webhook endpoint.

This PR add two new command line flags:

- _webhook-url_: Specifies the URL of the webhook service
- _certificates-namespace_: Allows to specify a namespace in which the webhook certificates are being stored. This is important because the namespace in which the webhook service is running may not be available in the targeting cluster. Therefore it needs to be specified manually.

**Release note**:

```feature user
Add command line flags "webhook-url" and "certificates-namespace" which allow to run the webhooks server in an external cluster.
```

### DIFF
--- a/pkg/utils/webhook/add.go
+++ b/pkg/utils/webhook/add.go
@@ -45,7 +45,7 @@ type Options struct {
 	// port of the service
 	ServicePort int32
 	// external service URL
-	ServiceURL string
+	WebhookURL string
 	// LabelSelector that is used to filter all resources handled by this webhook
 	ObjectSelector metav1.LabelSelector
 	// the resources that should be handled by this webhook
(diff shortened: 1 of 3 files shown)


pkg/utils/webhook/add.go
```diff
@@ -44,6 +44,8 @@ type Options struct {
 	ServiceNamespace string
 	// port of the service
 	ServicePort int32
+	// external service URL
+	ServiceURL string
```
schrodit, line 48: I think we should keep the naming and call it `WebhhokURL` otherwise the change from `webhookURL` int eh options to `serviceURL` in the webhook confuses

schrodit (changes requested):

pkg/utils/webhook/add.go
reshnm: Thank you. Done.

reshnm: Thank you for the review.
I have changed it to `WebhookURL`.

schrodit (approved):
/lgtm
/merge squash

PR #280, opened 2021-08-11
fixed the handling of the webhook-url command line flag

reshnm:

/area usability
/kind bug
/priority 3

When the `--webhook-url` flag was passed, the `URL` field in the `WebhookClientConfig` was not calculated correctly.
Additionally it is also no longer possible to derive the DNS names from the service name and namespace when using the `--webhook-url` flag.
With this change, the DNS name is directly derived from the webhook url.

**Release note**:

```

```

### DIFF
--- a/cmd/landscaper-webhooks-server/app/app.go
+++ b/cmd/landscaper-webhooks-server/app/app.go
@@ -127,20 +127,15 @@ func registerWebhooks(ctx context.Context,
 	// generate certificates
 	webhookServer.CertDir = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs")
 	var err error
-	var dnsNamesConfig webhook.DNSNamesConfig
+	var dnsNames []string
 	if len(wo.WebhookURL) != 0 {
-		conf := webhook.DNSWebhookURLConfig{}
-		err = conf.Set(wo.WebhookURL)
-		if err != nil {
+		if dnsNames, err = webhook.GetDNSNamesFromURL(wo.WebhookURL); err != nil {
 			return fmt.Errorf("unable to create webhook certificate configuration: %w", err)
 		}
-		dnsNamesConfig = &conf
 	} else {
-		conf := webhook.DNSNamespacedNameConfig{}
-		conf.Set(wo.ServiceName, wo.ServiceNamespace)
-		dnsNamesConfig = &conf
+		dnsNames = webhook.GeDNSNamesFromNamespacedName(wo.ServiceNamespace, wo.ServiceName)
 	}
-	wo.CABundle, err = webhook.GenerateCertificates(ctx, kubeClient, webhookServer.CertDir, o.webhook.certificatesNamespace, dnsNamesConfig)
+	wo.CABundle, err = webhook.GenerateCertificates(ctx, kubeClient, webhookServer.CertDir, o.webhook.certificatesNamespace, "landscaper-webhook", dnsNames)
 	if err != nil {
 		return fmt.Errorf("unable to generate webhook certificates: %w", err)
 	}
--- a/cmd/landscaper-webhooks-server/app/app.go
+++ b/cmd/landscaper-webhooks-server/app/app.go
@@ -132,7 +132,7 @@ func registerWebhooks(ctx context.Context,
 		conf := webhook.DNSWebhookURLConfig{}
 		err = conf.Set(wo.WebhookURL)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


achimweigel (approved):
/lgtm only a small typo

cmd/landscaper-webhooks-server/app/app.go
```diff
@@ -127,7 +127,20 @@ func registerWebhooks(ctx context.Context,
 	// generate certificates
 	webhookServer.CertDir = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs")
 	var err error
-	wo.CABundle, err = webhook.GenerateCertificates(ctx, kubeClient, webhookServer.CertDir, o.webhook.certificatesNamespace, wo.ServiceName)
+	var dnsNamesConfig webhook.DNSNamesConfig
+	if len(wo.WebhookURL) != 0 {
+		conf := webhook.DNSWebhookURLConfig{}
+		err = conf.Set(wo.WebhookURL)
+		if err != nil {
```
schrodit, line 134:
```suggestion
		if err = conf.Set(wo.WebhookURL); err != nil {
```

pkg/utils/webhook/certificates.go
```diff
@@ -26,9 +27,60 @@ const (
...
+type DNSNamespacedNameConfig struct {
+	config []string
+}
+
+// The DNSWebhookURLConfig derives the DNS names directly from the webhook url.
+type DNSWebhookURLConfig struct {
+	config []string
+}
+
+func (c *DNSNamespacedNameConfig) Set(name, namespace string) {
```
schrodit, line 45: docstring missing

pkg/utils/webhook/certificates.go
```diff
@@ -26,9 +27,60 @@ const (
...
+		fmt.Sprintf("%s.%s", name, namespace),
+		fmt.Sprintf("%s.%s.svc", name, namespace),
+	}
+}
+
+func (c *DNSNamespacedNameConfig) Get() []string {
+	return c.config
+}
+
+func (c *DNSWebhookURLConfig) Set(webhookURL string) error {
```
schrodit, line 57: docstring missing

pkg/utils/webhook/certificates.go
```diff
@@ -26,9 +27,60 @@ const (
...
+}
+
+func (c *DNSWebhookURLConfig) Get() []string {
+	return c.config
+}
+
 // GenerateCertificates generates the certificates that are required for a webhook. It returns the ca bundle, and it
 // stores the server certificate and key locally on the file system.
-func GenerateCertificates(ctx context.Context, kubeClient client.Client, certDir, namespace, name string) ([]byte, error) {
+func GenerateCertificates(ctx context.Context, kubeClient client.Client, certDir, namespace string, dnsNamesConfig DNSNamesConfig) ([]byte, error) {
```
schrodit, line 83: what is the benefit of the interface `DNSNamesConfig` instead of just passing a string array directly?

something like
```suggestion
func GenerateCertificates(ctx context.Context, kubeClient client.Client, certDir, namespace string, dnsNamesConfig []string) ([]byte, error) {
```

pkg/utils/webhook/certificates.go
```diff
@@ -117,8 +169,8 @@ func GenerateNewCAAndServerCert(namespace, name string, caConfig certificates.Ce
 	)
 
 	serverConfig := &certificates.CertificateSecretConfig{
-		CommonName:  name,
```
schrodit, line 120: we should keep that name as that function could be or is used to generally generate a ca server cert

schrodit (changes requested):

pkg/utils/webhook/certificates.go
reshnm: @schrodit The `CommonName` was previously derived from the service name. When we just have the webhook URL, how should it be set?

schrodit: I would have the value as another function parameter. With that we can configure it to use `landscaper-webhhok` but others have the chance to use whatever they want

reshnm: Ok, understood.

pkg/utils/webhook/certificates.go
reshnm: You are right, actually not much. I have removed the interface and pass the list directly.

schrodit (approved):
/lgtm

PR #284, opened 2021-08-12
Adapt json parser of type AnyJson

achimweigel:

/area deployers
/kind bug
/priority 3

If an AnyJson object is marshalled and the raw data are nil this results in the string "null". If objects like TemplateExecutor contain such an element they are not correctly handled when they are unmarshalled again, e.g. the in the unmarshalled object it is checkt if the raw data are empty, but in this case they contain "null". This PR resolves this problem by checking  during unmarshalling if the raw data contain "null" and resolves this to empty raw data.

Fixes #

**Release note**:

```bugfix developer
Landscaper is now able to parse json.RawMessage values that are `null`.
```


schrodit (changes requested):

schrodit:
@achimweigel I think you also have todo a `make revendor` ?

achimweigel:
we have done the revendoring

schrodit (approved):
/lgtm

PR #285, opened 2021-08-13
Adapt unmarshal AnyJSON

robertgraeff:

/area deployers
/kind bug
/priority 3

PR #284 has adjusted the unmarshaling of `AnyJSON`, so that "null" was unmarshaled into an `AnyJSON` whose `RawMessage` was a byte slice of length 0. 

The present PR modifies this and unmarshals "null" into an `AnyJSON` whose `RawMessage` is `nil`, so that this can be marshaled again.

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #288, opened 2021-08-17
fix the detection of target exports by a sibling

reshnm:

/area operations
/kind bug
/priority 3

Currently when a target is exported by sibling of an installation, this is not correctly detected when importing the target reference by another installation.
This PR fixes the the detection of target exports by installation siblings.

**Release note**:

``` bugfix user
A bug has been fixed that caused target export from installation to not be consumable by other installations.
```

### DIFF
--- a/pkg/landscaper/installations/imports/validation.go
+++ b/pkg/landscaper/installations/imports/validation.go
@@ -347,6 +347,7 @@ func (v *Validator) checkStateForParentImport(fldPath *field.Path, importName st
 	return nil
 }
 
+// The IsExportingDataFunc returns true if the passed sibling has a Data/Target export by the given name.
 type IsExportingDataFunc func(*installations.InstallationBase, string) bool
 
 func (v *Validator) checkStateForSiblingDataExport(ctx context.Context, fldPath *field.Path, siblingRef lsv1alpha1.ObjectReference, importName string) error {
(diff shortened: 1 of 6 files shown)


pkg/landscaper/installations/imports/validation.go
```diff
@@ -347,14 +347,30 @@ func (v *Validator) checkStateForParentImport(fldPath *field.Path, importName st
 	return nil
 }
 
+type IsExportingDataFunc func(*installations.InstallationBase, string) bool
```
schrodit, line 350: can you add a docstring explaining the function

schrodit (changes requested):
test looks good and also the implementation looks clean:)

some files are not formated. you can run `make format` to format all files.

pkg/landscaper/installations/imports/validation.go
reshnm: I can add a docstring.
However, since it is used only internally, should it be begin lowercase to avoid exporting?

schrodit: then I would propose to do both. so make it private and add a string what the boolean means

reshnm: done

schrodit (approved):
/lgtm

PR #292, opened 2021-08-18
add reconcile extension concept to deployer lib

Diaphteiros:

/area deployers
/kind enhancement
/priority 3

Requirement for the implementation of continuous reconciliation.

**Release note**:

```breaking developer
Extension hooks have been introduced to enable deployers to extend and modify the deployer library's default reconciliation flow. The `Deployer` interface has been expanded and its implementations now need an additional `ExtensionHooks()` method. Check out the documentation [here](docs/development/dep-lib-extension-hooks.md).
```


pkg/deployer/lib/controller.go
```diff
@@ -41,6 +42,10 @@ type Deployer interface {
 	ForceReconcile(ctx context.Context, di *lsv1alpha1.DeployItem, target *lsv1alpha1.Target) error
 	// Abort the deploy item progress.
 	Abort(ctx context.Context, di *lsv1alpha1.DeployItem, target *lsv1alpha1.Target) error
+	// AddExtensionHook adds an extension hook to this deployers hooks.
+	AddExtensionHook(hook extension.ReconcileExtensionHook)
```
Diaphteiros, line 46: @schrodit Is the `AddExtensionHook` method needed? Or are the only ones who add hooks the deployers itself, which can directly access the hook list and have no need for an add method?

pkg/deployer/lib/extension/state.go
```diff
@@ -0,0 +1,30 @@
...
+// All hooks will receive the pointer to the same state object and are allowed to modify it.
+type ReconciliationState struct {
+	// ReconcileResult can be used to end the reconcile loop.
+	ReconcileResult reconcile.Result
+
+	// If set to false, reconciliation will be aborted with returning ReconcileResult after the current execution.
+	ContinueReconcile bool
+
+	// GenericState can be used to transfer any state between hooks which doesn't fit in the other fields.
+	GenericState map[string]interface{}
```
schrodit, line 21: why do we need a generic state?
this state seems hard to debug and we do not really know who writes the data and who reads it

pkg/deployer/lib/extension/state.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// ReconciliationState represents the current state of the reconciliation.
+// All hooks will receive the pointer to the same state object and are allowed to modify it.
+type ReconciliationState struct {
+	// ReconcileResult can be used to end the reconcile loop.
+	ReconcileResult reconcile.Result
+
+	// If set to false, reconciliation will be aborted with returning ReconcileResult after the current execution.
+	ContinueReconcile bool
```
schrodit, line 18: we should rather flip the meaning so that the default is `false` and a new state does not need the new state constructor.

pkg/deployer/lib/extension/hook.go
```diff
@@ -0,0 +1,92 @@
...
+	EndHook                       HookType = "End"                       // called at the end of reconciliation
+)
+
+type ReconcileExtensionHook interface {
+	// Name returns the name of the extension hook.
+	// It is mainly used for logging purposes, but names are expected to be unique.
+	Name() string
+
+	// Type returns the hook type of the extension hook.
+	Type() HookType
```
schrodit, line 32: maybe type should be a number so that we could register multiple hooks like `StartHook|EndHook`.
As it is done in the fs.Mod package

pkg/deployer/lib/extension/hook.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Execute executes the hook.
+	// To influence the reconciliation loop, this function is expected to modify the given state accordingly.
+	Execute(*Operation, *ReconciliationState) error
```
schrodit, line 36: the call should include the current type.

pkg/deployer/lib/extension/operation.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
+)
+
+type Operation struct {
```
schrodit, line 16: not sure if we should just rather add all the values as function parameters.
Also I think we should only call the hook with the context parameters like the context, the deployitem and target.
other values should be available within the hook implementation itself

pkg/deployer/lib/extension/hook.go
Diaphteiros: I have refactored the extension hooks and it is now possibly to register a function for multiple hook types at once, which was the main point of your proposal.

I am not a fan of bit arithmetics because it is unintuitive and more difficult to understand and parse. I don't see any advantages in switching to numbers, except for a probably insignificant improvement in memory consumption, therefore I'd prefer to stick to strings. Feel free to discuss if you see additional advantages.

Diaphteiros:
Wohoo, the integration tests finally succeeded! @schrodit please take a look when you have time.

pkg/deployer/lib/extension/hook.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if hype == DuringResponsibilityCheckHook || hype == ShouldReconcileHook {
+		return AggregateHookResultsWithInvertedAbortPriority(hookRes...), nil
+	}
+	return AggregateHookResults(hookRes...), nil
+}
+
+// RegisterHook appends the given hook function to the list of hook functions for all given hook types.
+// It returns the ReconcileExtensionHooks object it is called on for chaining.
+func (hooks ReconcileExtensionHooks) RegisterHook(hook ReconcileExtensionHook, hypes ...HookType) ReconcileExtensionHooks {
```
reshnm, line 111: typo `hypes`

reshnm (changes requested):

pkg/deployer/lib/extension/hook.go
```diff
@@ -0,0 +1,126 @@
...
+type ReconcileExtensionHookSetup struct {
+	Hook      ReconcileExtensionHook
+	HookTypes []HookType
+}
+
+// ExecuteHooks executes all hooks of a given type in the order they are specified in.
+// The results of all executed hooks are aggregated using the AggregateHookResults function, except for
+//   DuringResponsibilityCheck and ShouldReconcile hooks, where AggregateHookResultsWithInvertedAbortPriority is used instead.
+// An error is returned if one of the hooks returns an error or if an unknown hook type is given.
+func (hooks ReconcileExtensionHooks) ExecuteHooks(ctx context.Context, log logr.Logger, di *lsv1alpha1.DeployItem, target *lsv1alpha1.Target, hype HookType) (*HookResult, error) {
```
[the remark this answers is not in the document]
Diaphteiros: This one is intended, as `type` is a keyword in go and can't be used.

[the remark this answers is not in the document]
reshnm: Ah, yes, makes sense.
It is a matter of taste, but `hype`seems strange to me as this is an english word with a totally different meaning.
Would something like `ht` be a better fit, since we are already using abbreviations like `di`f or deploy items?

docs/development/dep-lib-extension-hooks.md
```diff
@@ -0,0 +1,154 @@
...
+func (d *deployer) ExtensionHooks() extension.ReconcileExtensionHooks {
+	return d.hooks
+}
+```
+
+The extension library offers a helper method to register a new hook:
+```golang
+// RegisterHook appends the given hook function to the list of hook functions for all given hook types.
+// It returns the ReconcileExtensionHooks object it is called on for chaining.
+func (hooks ReconcileExtensionHooks) RegisterHook(hook ReconcileExtensionHook, hypes ...HookType) ReconcileExtensionHooks
```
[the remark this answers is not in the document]
Diaphteiros: This was intended, but let's discuss this: I used `hype` for the singular, as `type` is a reserved keyword in go (see below). Should I use `types` for the plural as this can be used or should I stick to the `hype` scheme and call it `hypes`?

pkg/deployer/lib/controller.go
```diff
@@ -144,60 +147,160 @@ func NewController(log logr.Logger,
...
+			hookRes = extension.AggregateHookResults(hookRes, tmpHookRes)
+			if hookRes.AbortReconcile {
+				return returnAndLogReconcileResult(logger, *hookRes), nil
+			}
+			if err := errHdl(ctx, c.delete(ctx, di, target)); err != nil {
+				return reconcile.Result{}, err
+			}
+		} else {
+			// default reconcile
+			logger.V(7).Info("handle deploy item reconcile")
```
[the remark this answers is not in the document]
Diaphteiros: It's intended. All the other cases are 'more interesting' than a standard reconcile, but I wanted to have a log output there too, if one wants to ensure that this case is actually executed while debugging. To not spam the logs, I decided for the higher verbosity here. Do you think it would be better to use `5` here too?

pkg/deployer/lib/extension/hook.go
Diaphteiros: see above

docs/development/dep-lib-extension-hooks.md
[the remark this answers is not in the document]
reshnm: See alos my answer below.
I think this is a matter of taste, and one should not argue about tastes ;-)
However, `hypes`or `hype` always looks like a typo for me.
I would rather go for an abbreviation like `ht`.
Final decision is up to you.

pkg/deployer/lib/controller.go
```diff
@@ -144,60 +147,160 @@ func NewController(log logr.Logger,
...
 
-	if !ShouldReconcile(deployItem) {
-		c.log.V(5).Info("aborting reconcile", "phase", deployItem.Status.Phase)
-		return reconcile.Result{}, nil
+	shouldReconcile = ShouldReconcile(di)
+	tmpHookRes, err = c.deployer.ExtensionHooks().ExecuteHooks(ctx, extensionLogger, di, target, extension.ShouldReconcileHook)
+	if err != nil {
+		return reconcile.Result{}, errHdl(ctx, err)
+	}
+	if !shouldReconcile {
```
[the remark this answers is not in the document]
Diaphteiros: It's especially needed if `ShouldReconcile` returns `false`, as the hook can overwrite this result to enforce a reconcile. But even if `ShouldReconcile` returns `true` - which cannot be turned into a `false` by the hook - we should still run them for the `Requeue` and `RequeueAfter` fields in the hook result.

pkg/deployer/lib/controller.go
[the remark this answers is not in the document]
reshnm: sounds reasonable

reshnm (approved):

pkg/deployer/lib/extension/hook.go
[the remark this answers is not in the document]
Diaphteiros: I like the idea. Will rename it to `ht`.

PR #293, opened 2021-08-18
fix possible deadlock in ResolveBlueprintFromBlobResolver

reshnm:

/area operations
/kind bug
/priority 3

If reading the blob fails, the pipe reader is not drained.
This will result in a deadlock inside the go routine while trying to write to the pipe.

**Release note**:

```

```

### DIFF
--- a/pkg/landscaper/blueprints/resolve.go
+++ b/pkg/landscaper/blueprints/resolve.go
@@ -131,6 +131,8 @@ func ResolveBlueprintFromBlobResolver(
 		}
 	}
 	blueprint, err := GetStore().Store(ctx, cd, resource, blobReader)
+	// drain the pipe reader, otherwise the error group wait may block indefinitely
+	_, _ = io.Copy(io.Discard, pr)
 	if err != nil {
 		if err2 := eg.Wait(); err2 != nil {
 			return nil, errorsutil.NewAggregate([]error{err, err2})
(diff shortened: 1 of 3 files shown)


pkg/landscaper/blueprints/resolve.go
```diff
@@ -131,6 +131,8 @@ func ResolveBlueprintFromBlobResolver(
 		}
 	}
 	blueprint, err := GetStore().Store(ctx, cd, resource, blobReader)
+	// drain the pipe reader, otherwise the error group wait may block indefinitely
+	_, _ = io.Copy(io.Discard, pr)
```
schrodit, line 135: ok I see that this can happen if the store does not need to read the blob.
This seems more like a quick workaround I would rather cancel the `blobReader.Resolve` as we otherwise always read from the remote.

I think we could give the blobreader a cancable context like (but we need to check if all blobresolver support the context)
```
eg, ctx := errgroup.WithContext(ctx)
remoteCtx, cancel := context.WithCancel(ctx)
	eg.Go(func() error {
		_, err := blobResolver.Resolve(remoteCtx, resource, pw)
		if err != nil {
			if err2 := pw.Close(); err2 != nil {
				return errorsutil.NewAggregate([]error{err, err2})
			}
			return fmt.Errorf("unable to resolve blueprint blob: %w", err)
		}
		return pw.Close()
	})
// then cancel if not needed anymore
cancel()
```

schrodit (changes requested):

pkg/landscaper/blueprints/resolve.go
schrodit: or another idea would be to (maybe even better) to handle the blueprint resolve in the `Store` method.
With that we would even call blobResolver.Reconcile if not needed.

schrodit (approved):

PR #298, opened 2021-08-31
Fix the handling of landscaper reference protocol schemes within JSON schemas

reshnm:

/area operations
/kind bug
/priority 3

The user should be able to use landscaper protocol schemes (`blueprint://`, `cd://`, `local://`) within a JSON schema.
This is currently not working as expected. Using a landscaper protocol scheme in a JSON schema results in the following error message:

`Get "cd://resources/my-schema": unsupported protocol scheme "cd":`

**Release note**:

```

```


schrodit (approved):
/lgtm

PR #299, opened 2021-08-31
add support for compressed jsonschema definitions

### LINKED ISSUE #297: compression type for JSON schema is ignored

/area operations
/kind bug
/priority 4

**What happened**:

Setting the media type of a JSON schema resource to a compressed type, `application/vnd.gardener.landscaper.jsonscheme.v1+json+gzip` will result in the following error:

```
{"level":"error","ts":"2021-08-30T08:40:06.728Z","logger":"setup.controllers.Installations.controller.installation","msg":"Reconciler error","reconciler group":"landscaper.gardener.cloud","reconciler kind":"Installation","name":"virtual-landscaper-m5r8q","namespace":"default","error":"invalid character '\\x1f' looking for beginning of value: "}
```

It seems that the media type compression is not handled for JSON schemas.

The compressed JSON schema should be decompressed before being interpreted or the error message should state that compression is not supported for JSON schemas.

Set media type of JSON schema to `application/vnd.gardener.landscaper.jsonscheme.v1+json+gzip` and set `compress: true`.

**Environment**:

- Landscaper version: v0.12.0
- Kubernetes version (use `kubectl version`): N/A
- Others:

schrodit:

/kind usability
/priority 3

Fixes https://github.com/gardener/landscaper/issues/297

**Release note**:

```other operator
Support for compressed jsonschema definitions has been added
```

### DIFF
--- a/pkg/landscaper/jsonschema/loader.go
+++ b/pkg/landscaper/jsonschema/loader.go
@@ -6,9 +6,11 @@ package jsonschema
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/url"
 	"path/filepath"
 
@@ -18,6 +20,8 @@ import (
 	"github.com/xeipuuv/gojsonreference"
 	"github.com/xeipuuv/gojsonschema"
 	"sigs.k8s.io/yaml"
+
+	"github.com/gardener/landscaper/apis/mediatype"
 
 	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
 	"github.com/gardener/landscaper/pkg/landscaper/registry/components/cdutils"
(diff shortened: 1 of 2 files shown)


pkg/landscaper/jsonschema/loader.go
```diff
@@ -186,10 +190,30 @@ func (l *Loader) loadComponentDescriptorReference(refURL *url.URL) ([]byte, erro
...
+	info, err := blobResolver.Resolve(ctx, res, &JSONSchemaBuf)
 	if err != nil {
 		return nil, fmt.Errorf("unable to fetch jsonschema for '%s': %w", refURL.String(), err)
 	}
 
+	mt, err := mediatype.Parse(info.MediaType)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse media type %q: %w", info.MediaType, err)
+	}
+	if mt.Type != mediatype.JSONSchemaArtifactsMediaTypeV1 {
```
reshnm, line 202: With this change it seems that we would drop support for `JSONSchemaArtifactsMediaTypeV0_`.
Should this be added to the release notes?

reshnm:
@schrodit This change is about comporessed JSON Schema, not component descriptor. You might want to update the title/commit?

pkg/landscaper/jsonschema/loader.go
schrodit: no its not dropped as `JSONSchemaArtifactsMediaTypeV0` is automatically converted to `JSONSchemaArtifactsMediaTypeV1` here: https://github.com/gardener/landscaper/blob/master/apis/mediatype/constants.go#L51

reshnm: 👍

reshnm (approved):

PR #304, opened 2021-09-03
[WIP] Context

### LINKED ISSUE #241: Namespaced Context

/area security
/kind epic
/priority 4

Currently, the Landscaper is not really able to handle multiple tenants.
One reason is that the default repository context and all repository secrets are shared.

It is already possible to set the repository and pull secrets on installations.
With that the pull secrets reside in the tenants namespace and the repository can differ.

But to keep the original intention of having one repository context and the installation should not care about it, we could introduce a new `Context` object.

A context object would contain any information that is currently shared but that new object would be namespaced.
Also there will be one `default` context in a namespace which all installation default to, but it could be also possible to add multiple contexts and reference them in the installation.

So a simple context object would look as follows:
```yaml
apiVersion: landcaper.gardener.cloud/v1alpha1
kind: Context
metadata:
  name: default
  namespace: default
spec:
  repositoryContext: # default repository context
    type: ociRegistry
    baseUrl: my-registry
  registryPullSecrets: # reference pull secrets in the same namespace.
  - my-pull-secret
```

Another advantage of such an object would be that we could enhance the context easily with other credentials or configuration that should be shared.
For example we could have a shared object that contains vault access information that can then be used by the landscaper and all deployers.
Another example is that specific deployer configuration that should not be part of the deploy item could be configured. (Here this would be again access credentials for repositories or other).

Something like that could then look like:
```yaml
apiVersion: landcaper.gardener.cloud/v1alpha1
kind: Context
metadata:
  name: default
  namespace: default
spec:
  ...
  secretStorageConfig:
     apiVersion: my-secret-store.domain.com/v1alpha1
     kind: Configuration
     endpoint: my.secret-store.com/some-namespace
     access:
        secretRef:
          name: abc

  deployerConfig:
  - type: helm.landscaper.gardener.cloud # unique identifier of the deployer type. same as in depoloy items
     config: # here comes some deployer specific config
```

schrodit:

/area landscaper
/kind api-change
/priority 3

Fixes #241 

**Release note**:

```other operator

```


pkg/landscaper/installations/scope.go
```diff
@@ -0,0 +1,130 @@
...
+
+// SetInstallationScope determines the current context and updates the operation context.
+func (o *Operation) SetInstallationScope(ctx context.Context) error {
+	newCtx, err := o.DetermineInstallationScope(ctx)
+	if err != nil {
+		return err
+	}
+	o.scope = *newCtx
+	return nil
+}
```
achimweigel, line 40: The usage of context and scope is a little bit confusing here. What is the difference?

achimweigel (approved):
/lgtm 
just two small comments

schrodit:
split in multiple PR's
Following up with https://github.com/gardener/landscaper/pull/325

PR #305, opened 2021-09-07
fix defaulting for container deploy items

Diaphteiros:

/area container-deployer
/kind bug
/priority 3

**Release note**:

```bugfix user
Fixed a bug regarding the defaulting of the container deployer which could cause it to generate invalid pod manifests.
```


achimweigel (approved):
/lgtm

pkg/deployer/container/container.go
```diff
@@ -105,6 +105,10 @@ func New(log logr.Logger,
 }
 
 func applyDefaults(config *containerv1alpha1.Configuration, providerConfig *containerv1alpha1.ProviderConfiguration) {
+	// default configuration
```
schrodit, line 108: nit
```suggestion
```

PR #306, opened 2021-09-08
added new landscaper controller option "--landscaper-kubeconfig"

reshnm:

/area usability
/kind enhancement
/priority 3

This PR adds a new landscaper option `--landscaper-kubeconfig`.
With this option the landscaper controller can run in a hosting cluster but operate on landscaper resources in a different cluster.
This PR also adds new kubeconfig properties to the landscaper helm chart.
When this properties are set for the landscaper controller/webhooks server, they will automatically use this new feature.

**Release note**:

```feature user
Add new landscaper controller command line option "--landscaper-kubeconfig".
This option allows the landscaper controller to run inside a hosting cluster, but operate on landscaper resources in a different cluster.
```

### DIFF
--- a/charts/landscaper/templates/cluster-kubeconfig-secret.yaml
+++ b/charts/landscaper/templates/cluster-kubeconfig-secret.yaml
@@ -1,0 +1,26 @@
+# SPDX-FileCopyrightText: 2020 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if .Values.controller.clusterKubeconfig.kubeconfig }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "landscaper.fullname" . }}-controller-cluster-kubeconfig
+  labels:
+    {{- include "landscaper.labels" . | nindent 4 }}
+data:
+  kubeconfig: {{ .Values.controller.clusterKubeconfig.kubeconfig | b64enc }}
+{{- end }}
+{{- if .Values.webhooksServer.clusterKubeconfig.kubeconfig }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "landscaper.fullname" . }}-webhooks-cluster-kubeconfig
+  labels:
+    {{- include "landscaper.labels" . | nindent 4 }}
+data:
+  kubeconfig: {{ .Values.webhooksServer.clusterKubeconfig.kubeconfig | b64enc }}
+{{- end }}
--- a/charts/landscaper/templates/cluster-kubeconfig-secret.yaml
+++ b/charts/landscaper/templates/cluster-kubeconfig-secret.yaml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-{{- if .Values.controller.clusterKubeconfig.kubeconfig }}
+{{- if .Values.controller.landscaperKubeconfig.kubeconfig }}
 ---
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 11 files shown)


charts/landscaper/values.yaml
```diff
@@ -59,6 +59,15 @@ image: {}
 #  tag: ""
 
 controller:
+  # Allows to target a different cluster than the hosting cluster of the controller instance.
+  # Provide the kubeconfig to access the remote Landscaper cluster here (inline or via secretRef).
+  # When providing a secretRef, see ./templates/cluster-kubeconfig-secret.yaml for the correct secret format.
+  # If no value is provided at all, the controller will default to the in-cluster kubeconfig.
+  clusterKubeconfig: {}
```
schrodit, line 66: not sure if we should rather name it something like `landscaperKubeconfig` or `lsKubeconfig`.
only `clusterKubeconfig` might be confusing

charts/landscaper/templates/cluster-kubeconfig-secret.yaml
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: 2020 SAP SE or an SAP affiliate company and Gardener contributors
```
schrodit, line 1: The landscaper helm chart currently also contains rbac rules for creating crd's, etc.. that are needed on the remote cluster.
But with that change these rbac rules and service account would get deployed on the host cluster which is not really what we want.

So I suggest to not deploy the rbac rules and service accounts when a remote landscaper cluster is specified.
or maybe move the definition to a dedicated chart as it is done in the gardener with their `runtime` and `application` charts

schrodit (changes requested):

charts/landscaper/values.yaml
reshnm: done

charts/landscaper/templates/cluster-kubeconfig-secret.yaml
reshnm: done

charts/landscaper/templates/deployment-webhooks.yaml
```diff
@@ -28,8 +28,13 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       {{- if .Values.serviceAccount.create }}
+      {{- if not .Values.webhooksServer.landscaperKubeconfig }}
       serviceAccountName: {{ include "landscaper.webhooks.serviceAccountName" . }}
       {{- end }}
+      {{- end }}
+      {{- if .Values.webhooksServer.landscaperKubeconfig }}
```
schrodit, line 35:
```suggestion
      {{- else }}
```

cmd/landscaper-controller/app/app.go
```diff
@@ -77,11 +81,32 @@ func (o *Options) run(ctx context.Context) error {
...
+	if len(o.landscaperKubeconfigPath) > 0 {
+		data, err := ioutil.ReadFile(o.landscaperKubeconfigPath)
+		if err != nil {
+			return fmt.Errorf("unable to read landscaper kubeconfig from %s: %w", o.landscaperKubeconfigPath, err)
+		}
+		client, err := clientcmd.NewClientConfigFromBytes(data)
+		if err != nil {
+			return fmt.Errorf("unable to build landscaper cluster client from %s: %w", o.landscaperKubeconfigPath, err)
+		}
+		hostConfig, err := client.ClientConfig()
```
schrodit, line 99: nit: that we do not confuse the host cluster with the landscaper cluster
```suggestion
		lsConfig, err := client.ClientConfig()
```

schrodit (changes requested):

charts/landscaper/templates/deployment-webhooks.yaml
reshnm: I even think that checking for `.Values.serviceAccount.create` is not correct in this case.
We also do not check for this value in the controller deployment.
When `.Values.webhooksServer.landscaperKubeconfig` is not set, the deployment needs to have a serviceAccount bound with the correct roles attached, despite of  `.Values.serviceAccount.create`.

cmd/landscaper-controller/app/app.go
reshnm: done

schrodit (approved):
/lgtm

PR #311, opened 2021-09-13
add unrecoverable reconcile test

### LINKED ISSUE #310: Enhance Deployer Provider Configuration parsing

/area deployers
/kind enhancement
/priority 4

Currently deployers return a default landscaper error when the provider configuration cannot be parsed or contain structural errors.
This means that the deploy item goes into failed state only after 5 minutes.
Such error are unrecoverable so we should directly go into an error state with a specific error code.

Maybe we could also handle this either with a specific error code or with an optional `unrecoverable` attribut in the error object.

schrodit:

/area deployers
/kind test
/priority 3

Adds a test for unrecoverable error that are not retried

Fixes #310 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/apis/deployer/container/v1alpha1/validation/validation.go
+++ b/apis/deployer/container/v1alpha1/validation/validation.go
@@ -5,10 +5,24 @@
 package validation
 
 import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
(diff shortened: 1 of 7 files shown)


achimweigel (approved):
/lgtm

PR #314, opened 2021-09-14
Improve integration tests

Diaphteiros:

/area testing
/kind technical-debt
/priority 3

This improves two integration tests:
1. One of the deployer management tests, which checks for failed installations now only observes installations which were created during the test. This prevents the test from failing if there are already failed installations in the landscaper namespace.
2. The phase propagation test now tries to update the status multiple times, in case it fails. This prevents flakeyness due to update conflicts.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/integration/deployers/management/dm_tests.go
+++ b/test/integration/deployers/management/dm_tests.go
@@ -206,6 +206,10 @@ func DeployerManagementTests(f *framework.Framework) {
 			instList := &lsv1alpha1.InstallationList{}
 			testutil.ExpectNoError(f.Client.List(ctx, instList, client.InNamespace(f.LsNamespace)))
 			numOfInstallations := len(instList.Items)
+			previousInstallations := sets.String{}
+			for _, inst := range instList.Items {
+				previousInstallations.Insert(inst.Name)
+			}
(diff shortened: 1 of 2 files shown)


test/integration/deployers/management/dm_tests.go
```diff
@@ -232,17 +236,23 @@ func DeployerManagementTests(f *framework.Framework) {
...
+					}
+				}
+				newInstallationCount := len(newInstallations)
+				if newInstallationCount != numOfEnvironments {
+					err := fmt.Errorf("expected %d new installations but found %d", numOfEnvironments, newInstallationCount)
 					f.TestLog().Logln(err.Error())
 					return err
 				}
-				// expect that all installations are healthy
+				// expect that all installations are healthy - TODO check only for created installation(s)
```
schrodit, line 251: that todo is done?

schrodit (commented):
otherwise lgtm

test/integration/deployers/management/dm_tests.go
Diaphteiros: you're right

schrodit (approved):

PR #317, opened 2021-09-15
fix landscaper deletion lifecycle

### LINKED ISSUE #230: Installation is never deleted

/area robustness
/kind bug
/priority 3

**What happened**:
Suppose an installation fails, because its blueprint cannot be resolved:

```yaml
  lastError:
    ...
    message: 'unable to find blueprint blueprint in component descriptor: NotFound'
    operation: InitPrerequisites
    reason: ResolveBlueprint
  observedGeneration: 0
  phase: Failed
```

I run `kubectl delete installation ...` for the installation. The installation gets a deletion timestamp, but the landscaper does not delete it. After an hour it is still there.

The installation should be deleted.

I am aware that checks are done before a deletion, and that they might need the blueprint. But in this case the failure of the installation occurred so early that no execution and deploy item was created. Shouldn't it be possible to delete the installation? 

**Environment**:
The current version of the landscaper in the master branch.

schrodit:

/area operations
/kind cleanup
/priority 3

- Fixes an issue that installations that never had any subinstallations nor blueprint can be deleted even if some configuration is erroneous.

- The deletion flow has been enhanced so that now changes in the installation are propagated to the execution and subinstallations.
Before this was not possible so one was unable to fix anything after deletion was triggered.

- The check if a installation is ready to be executed is now done before all external resources are fetched. This should decrease the network and increase the performace for most of the reconcile loops.

Fixes #230 

**Release note**:

```bugfix user
A bug has been fixed that caused the installation to stuck in deletion if the installation is erroneous.
Updated configuration is now correctly propagated during deletion and it is possible to set the `force-reconcile` annotation.

Installations that never had a successful execution (never have deployed an execution nor an installation) are now directly deleted.
```

### DIFF
--- a/pkg/landscaper/controllers/installations/controller.go
+++ b/pkg/landscaper/controllers/installations/controller.go
@@ -88,6 +88,7 @@ type Controller struct {
 
 func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	logger := c.Log().WithValues("installation", req.NamespacedName.String())
+	ctx = logr.NewContext(ctx, logger)
 	logger.V(5).Info("reconcile", "resource", req.NamespacedName)
 
 	inst := &lsv1alpha1.Installation{}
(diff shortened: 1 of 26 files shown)


reshnm (approved):
I can't comment on every detail of this PR.
But overall it looks very good to me.

I also did some tests locally and it worked as expected.

PR #319, opened 2021-09-15
remove undefined exports from exportExecution result

schrodit:

/kind bug
/priority 3

Exports that are not defined by the blueprint are now removed from the templating result.

**Release note**:

```breaking user
Exports that are not defined by the blueprint are now removed from the templating result.
Be aware that this might break blueprint that export data which is not defined.
```


reshnm (approved):
/lgtm

Diaphteiros (approved):

robertgraeff (approved):

PR #321, opened 2021-09-16
Add k8s resources creation and updating order

### LINKED ISSUE #312: Add resource dependencies in helm and manifest deployer

/area manifest-deployer
/area helm-deployer
/kind enhancement
/priority 3

Currently the helm charts are rendered and then templates are applied as rendered by helm.
It can happen that inside the helm there are certain dependencies towards other resource like namespaces have to be created before other resources can be created inside that namespace. (reported in https://github.com/gardener/landscaper/issues/309)

Therefore we either have some sorting or dependency management or try to create all resources in parallel and retry.

With that ticket we should also have a sorted list of managed resources to not retrigger updated all the time.
And we should check with the apiresources for namespaced and non-namespaced resources.

**Why is this needed**:
This is needed to apply helm charts a manifests that depend on each other.

schrodit:

/area manifest-deployer
/area helm-deployer
/kind bug
/priority 3

Fixes #312

**Release note**:

```bugfix user
Adds dependencies between kubernetes resources in the helm and manifest deployer so resources can be deployed that requires other resources.
For example namespaces are now deployed before any other namespaced resources.
```


reshnm (approved):
/lgtm

PR #324, opened 2021-09-20
simplfied the usage of the teststate helper struct

schrodit:

/kind cleanup
/priority 3

The state interface of the test helper has been refactored to now store the client.
With that the interface has been simplified as the client does not have to passed in every call.

Fixes #

**Release note**:

```other operator
NONE
```


Diaphteiros (approved):
/lgtm

PR #325, opened 2021-09-21
Add api-definitions for the context

schrodit:

/kind api-change
/priority 3

Adds the api definition for the new context.

Currently the context only contains the repository context and optional pullrequest secrets.
But it can be enhanced if needed as described in the ticket #241 

is part of #241

**Release note**:

```other operator
NONE
```


apis/core/types_installation.go
```diff
@@ -40,6 +40,10 @@ type Installation struct {
 
 // InstallationSpec defines a component installation.
 type InstallationSpec struct {
+	// Context defines the current context of the installation.
+	// +optional
+	Context string `json:"context,omitempty"`
```
Diaphteiros, line 45: Shouldn't that be a reference to a context resource, consisting of name and namespace?
If a context resource is namespaced but referenced only by name, the namespace would need to be defaulted or we have to ensure that there exists exactly one context with that name across all namespaces (which is strongly counter-intuitive regarding the concept of namespaces).

Diaphteiros (changes requested):
Please explain why the namespaced `Context` resource is referenced via a simple string and not via an object reference with name and namespace.
Also, there is no documentation for the Context, but since this only introduces the API, it should be fine to add the documentation together with the actual functionality.

apis/core/types_installation.go
Diaphteiros: OK, now that I think about it, it doesn't make any sense to allow referencing contexts in arbitrary namespaces, as this would grant users access to other tenants' registries.

Diaphteiros: So the namespace of the context is always the same as the one where the installation is in? And there is an exception for the `default` context which lies in the `default` namespace?

achimweigel (approved):
/lgtm

But I have the feeling that I reviewed a similar PR some days ago?

Diaphteiros:
> But I have the feeling that I reviewed a similar PR some days ago?
The contents of this PR are also part of #304 ;-)

apis/core/types_installation.go
schrodit: > So the namespace of the context is always the same as the one where the installation is in? And there is an exception for the default context which lies in the default namespace?

yes it is required that the context is in the same namespace as the installation.
Therefore no namespace is needed.

If no reference is given, the default context is used. There is not just one default context in the default namespace.
The concept is rather similar to the default service account. Every namespace has their own default context.

schrodit:
> Also, there is no documentation for the Context, but since this only introduces the API, it should be fine to add the documentation together with the actual functionality.

This PR only introduces the api definitions.
So nothing can be currently done with it.
I will add the documentation with the implementation PR.

Diaphteiros (approved):
/lgtm

PR #330, opened 2021-09-28
add example for the preserveDir option

reshnm:

/area documentation
/kind enhancement
/priority 4

Specifying the correct path for a helm resource is a common pitfall.
Improve the docuementation a bit, so that it is more clear that the parent path to the helm chart has to be specified.
Also explain the `preserveDir` option.

**Release note**:

```
NONE
```

### DIFF
--- a/docs/tutorials/02-local-simple-blueprint.md
+++ b/docs/tutorials/02-local-simple-blueprint.md
@@ -85,8 +85,27 @@ version: v0.1.0
 relation: external
 input:
   type: "dir" # type dir automatically creates a tar from the given dir.
-  path: /tmp/chart # path to the downloaded helm chart
+  path: /tmp/chart # parent path to the downloaded helm chart
   compress: true # compresses the tar/blob using gzip.
+  # Media type automatically defaults to "application/gzip" for compressed=true
+  # mediaType: "application/gzip"
+...
+```
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/02-local-simple-blueprint.md
```diff
@@ -85,13 +85,31 @@ version: v0.1.0
...
-  path: /tmp/chart # path to the downloaded helm chart
+  path: /tmp/chart # parent path to the downloaded helm chart
   compress: true # compresses the tar/blob using gzip.
   # Media type automatically defaults to "application/gzip" for compressed=true
   # mediaType: "application/gzip"
 ...
 ```
 
+:warning: In the above example, the path to the helm chart points to the parent directory of `/tmp/chart/nginx-ingress`. 
+If the fully qualified path to the helm chart should be used, the option `preserveDir` has to be set to `true`.
```
schrodit, line 96: maybe we should also add an explanation why this is needed -> helm structure enforces a parent directory

schrodit (changes requested):

docs/tutorials/02-local-simple-blueprint.md
reshnm: 👍 What about now?

schrodit (approved):

PR #331, opened 2021-09-28
Continuous reconcile extension

Diaphteiros:

/area deployers
/kind enhancement
/priority 3

It adds a continuous reconcile extension to the deployer library which can be used by deployers to implement continuous reconciliation. This is demonstrated as an example in the mock deployer.

Part of #239 

**Release note**:

```feature developer
A deployer library extension which enables continuous reconciliation of deploy items has been implemented. See [here](docs/development/deployer-extensions.md#continuous-reconcile-extension) for the documentation.
```


pkg/deployer/lib/continuousreconcile/extension.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// ContinuousReconcileExtension returns an extension hook function which can handle continuous reconciliation.
+// It is meant to be used as a ShouldReconcile hook and might yield unexpected results if used with another hook handle.
+// The function which it takes as an argument is expected to take a time and return the time when the deploy item should be scheduled for the next automatic reconciliation.
+//   It should return nil if continuous reconciliation is not configured for the deploy item.
+func ContinuousReconcileExtension(nextReconcile func(context.Context, time.Time, *lsv1alpha1.DeployItem) (*time.Time, error)) extension.ReconcileExtensionHook {
+	return func(ctx context.Context, log logr.Logger, di *lsv1alpha1.DeployItem, target *lsv1alpha1.Target, hype extension.HookType) (*extension.HookResult, error) {
+		logger := log.WithName("continuousReconcileExtension")
+		logger.V(7).Info("execute")
```
schrodit, line 27: this logging output seems not really helpful.
I think we should provide more context

pkg/deployer/lib/continuousreconcile/extension.go
```diff
@@ -0,0 +1,91 @@
...
+		if di == nil {
```
schrodit, line 28: how can this happen?

pkg/deployer/lib/continuousreconcile/extension.go
```diff
@@ -0,0 +1,91 @@
...
+			logger.V(7).Info("no continuous reconcile specified")
+			return nil, nil
+		}
+		next := nextRaw.Round(time.Second)
+		now := time.Now().Round(time.Second)
+		res := &extension.HookResult{}
+		if next.After(now) {
+			// reconcile is not yet due
+			// since this is a ShouldReconcile hook, we can set AbortReconcile to false and the deploy item will still be reconciled if necessary (e.g. due to spec changes)
+			res.AbortReconcile = true
```
schrodit, line 45: can't we just directly return here as we are not due?

schrodit (changes requested):

apis/deployer/utils/continuousreconcile/doc.go
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors.
```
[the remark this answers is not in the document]
Diaphteiros: OK, I will remove the versioning.

apis/deployer/utils/continuousreconcile/types.go
```diff
@@ -0,0 +1,44 @@
...
+	// +optional
+	Cron string `json:"cron,omitempty"`
+
+	// Every specifies a delay after which the reconcile should happen.
+	// Either Cron or Every has to be specified.
+	// +optional
+	Every *lscore.Duration `json:"every,omitempty"`
+}
+
+func (crs ContinuousReconcileSpec) String() string {
```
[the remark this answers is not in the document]
Diaphteiros: Because the validation will print the struct if it is invalid and since the struct contains a pointer whose address changes with every run, the corresponding deploy item will be reconciled endlessly (because the error message changes every time).

apis/deployer/utils/continuousreconcile/types.go
```diff
@@ -0,0 +1,44 @@
...
+	sb.WriteString("{")
+	if len(crs.Cron) != 0 {
+		sb.WriteString("Cron: \"")
+		sb.WriteString(crs.Cron)
+		sb.WriteString("\"")
+		if crs.Every != nil && crs.Every.Duration != 0 {
+			sb.WriteString(", ")
+		}
+	}
+	if crs.Every != nil && crs.Every.Duration != 0 {
```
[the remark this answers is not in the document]
Diaphteiros: The stringer method will be called by the validation for creating the error message, so it could be the case that both fields are set (in which case the spec would be invalid).
The reason for the 'complexity' is the pointer topic explained above.

hack/tools.go
```diff
@@ -1,3 +1,4 @@
+//go:build tools
```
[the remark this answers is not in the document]
Diaphteiros: `make format` with golang 1.17

pkg/deployer/lib/continuousreconcile/extension.go
Diaphteiros: This is meant to log that the execution of the continuousReconcileExtension has started, similar to the 'reconcile' log at the beginning of reconciliation. What context would you like to add?

pkg/deployer/lib/continuousreconcile/extension.go
Diaphteiros: It's more of a safety measure, it can't happen unless someone uses the extension in a wrong way, I guess.

pkg/deployer/lib/continuousreconcile/extension.go
Diaphteiros: No, we still need to compute the time when the continuous reconciliation is expected to be due and return a corresponding `RequeueAfter` value ...

hack/tools.go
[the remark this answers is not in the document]
schrodit: hmm interesting.
I also use 1.17 but when I run `make format` the header does not change. Maybe we should investigate why this has changed

pkg/deployer/lib/continuousreconcile/extension.go
schrodit: ok but if it happens we should return an error then?
otherwise it would just silently not be called.
Maybe even a panic is better

Diaphteiros: Good point 🤔

Diaphteiros: I changed it to return an error in this case.

pkg/deployer/lib/continuousreconcile/extension.go
schrodit: ah I see that you added a `WithName` so this is fine.
Although we might want to have development on how the verbosity levels should be used

pkg/deployer/lib/continuousreconcile/extension.go
Diaphteiros: Update: I changed it to a panic.

apis/go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/go-openapi/spec v0.19.5
 	github.com/onsi/ginkgo v1.14.2
 	github.com/onsi/gomega v1.10.4
+	github.com/robfig/cron/v3 v3.0.1
```
schrodit, line 11: can we remove the dependency towards the cron in the apispec?
I think we only really need it in the landscaper/deployer code.

schrodit (changes requested):

apis/go.mod
Diaphteiros: It's currently required by the validation of the ContinuousReconcileSpec: https://github.com/Diaphteiros/landscaper/blob/continuous-reconcile/apis/deployer/utils/continuousreconcile/validation/validation.go#L35 ...

schrodit: ok then let's keep. I think the validation more useful than having one less dependency.

schrodit (approved):

PR #333, opened 2021-09-30
pin ingress-nginx chart version to v3.29.0

reshnm:

/area documentation
/kind enhancement
/priority 3

In the tutorial we are pulling the latest version of the ingress-nginx helm chart.
However, newer versions (4.0.0+) have a breaking change in the chart values which makes the tutorial fail.
We should pull a specific version of the chart which is known to be compatible with our tutorial.

The version is now pinned to v3.29.0 which is also used in our integration tests.

**Release note**:

```other operator
NONE
```


schrodit (approved):
/lgtm

PR #334, opened 2021-09-30
Context

### LINKED ISSUE #241: Namespaced Context

schrodit:

/area scalability
/kind api-change
/priority 3

This PR adds a new context object that contains configuration that is shared across installations in a namespace.
This object enables multi tenant capabilities as previously defined global config can now be defined namespaced.
An Example for such a configuration are registry pull secrets or the repository context.

Everything is implemented backwards compatible so all installations should work as before.
This backwards-compatibility is archived by deploying a default context in each namespace that is automatically referenced.

Fixes https://github.com/gardener/landscaper/issues/241

**Release note**:

```feature user
It is now possible to define namespaced context object that contain previous global configuration like the repository context or registry pull secrets.
It enables the Landscaper to handle multiple tenants separated by namespace.

The Context feature is backwards-compatible so all installations should work as before.
This backwards-compatibility is archived by deploying a default context in each namespace that is automatically referenced by installations.

See the `Context` and `Installation` documentation for more details.
```


schrodit:
@reshnm do I need to adjust the landscaper blueprint config jsonschema definitions?

docs/usage/Installations.md
```diff
@@ -122,6 +127,23 @@ status:
       namespace: default
 ```
 
+## Context
+
+The context is a configuration resource containing shared configuration for installations.
+This config can contain the repository context, registry pull secrets, etc. . For detailed documentation see [./Context.md](./Context.md).
+
+Installations *MUST* reference a context by its name in `.spec.context` (it is defaulted to the `default` context if not defined).
+The context must be present in the same namespace as the installtion.
```
reshnm, line 136: typo `installtion`

docs/usage/Installations.md
```diff
@@ -122,6 +127,23 @@ status:
       namespace: default
 ```
 
+## Context
+
+The context is a configuration resource containing shared configuration for installations.
+This config can contain the repository context, registry pull secrets, etc. . For detailed documentation see [./Context.md](./Context.md).
+
+Installations *MUST* reference a context by its name in `.spec.context` (it is defaulted to the `default` context if not defined).
```
reshnm, line 135: If a `context` key is required, we loose backwards-compatibility or am I reading this wrong?
Will it be allowed to ommit the `spec.context`key, which then would default to the `default` context,
or do we expect at least a `context: ""`?

examples/00-Landscaper-Configuration.yaml
```diff
@@ -5,6 +5,15 @@ repositoryContext: # define the default repository context for installations
...
   baseUrl: "myregistry.com/components"
 
+controllers:
+  context:
+    config:
+      disableDefault: false
+      repositoryContext: # define the default repository context for installations
+        type: ociRegistry
+        baseUrl: "myregistry.com/components"
+
```
reshnm, line 15: I'm a little bit confused here.
Is it

```yaml
controllers:
  context:
    config:
      default:
        disable: false
```

or

```yaml
controllers:
  context:
    config:
      disableDefault: false
```

Are both schemes allowed?

docs/usage/Context.md
```diff
@@ -0,0 +1,55 @@
...
+    config:
+      default:
+        disable: false
+        excludeNamespaces: # optional
+        - kube-system
+        - ls-system
+        repositoryContext: # define the default repository context for installations
+          type: ociRegistry
+          baseUrl: "myregistry.com/components"
+```
```
reshnm, line 55: I got some questions one my mind while reading this.
The landscaper will create a default context in each namespace except it is excluded in `excludeNamespaces`.
What about non-default contexts? I can specify them in the landscaper configuration, right? Are the also created by the landscaper or do they have to be created manually? In which namespace are they living or do they have  to be created manually by the user?

pkg/landscaper/installations/context.go
```diff
@@ -52,29 +67,69 @@ func (o *Operation) DetermineInstallationContext(ctx context.Context) (*Context,
...
-		Parent: parent,
+		Parent: CreateInternalInstallationBase(parentInst),
 		// siblings are all encompassed installation of the parent installation
 		Siblings: CreateInternalInstallationBases(siblingInstallations...),
+		External: externalCtx,
 	}, nil
 }
 
+// ExternalContext defines the internal context with additional enhanced context information.
+type ExternalContext struct {
```
reshnm, line 89: I'm a bit lost about what makes a context `internal`or `external` ...
External tho whom?

reshnm (changes requested):

reshnm:
> @reshnm do I need to adjust the landscaper blueprint config jsonschema definitions?

I think it makes sense the reflect this also in the schema definititions and the landscaper blueprint, so that we can pass the context configuration in the installation.

docs/usage/Installations.md
schrodit: you're right the wording is a bit confusing.
Id the context is omitted the key is defaulted to `"default"`

examples/00-Landscaper-Configuration.yaml
schrodit: missed that one during some refactoring.
the first one is correct

docs/usage/Context.md
schrodit: non-defaults contexts can be manually created by the users using e.g. `kubectl apply`.
There are alos just a resource like installations. so they are namespaced CR's and teh namespaced is spcified during creation

schrodit:
> I think it makes sense the reflect this also in the schema definititions and the landscaper blueprint, so that we can pass the context configuration in the installation.

ok then should should maybe think of some automatic generation of the jsonschema. WDYT?

reshnm:
> > I think it makes sense the reflect this also in the schema definititions and the landscaper blueprint, so that we can pass the context configuration in the installation.
> 
> ok then should should maybe think of some automatic generation of the jsonschema. WDYT?

I think that makes sense. Could probably also be done in a seperate PR?

docs/usage/Context.md
reshnm: But why would we then add them to the landscaper configuration?
My impression was, I could specify them like this:

```yaml
controllers:
  context:
    config:
     myContext:
       repositoryContext:
         type: "ociRegistry"
         baseUrl: ""

registryConfig:
  secrets:
    myContext: {}
```

Or am I wrong?

schrodit:
> I think that makes sense. Could probably also be done in a seperate PR?
yes otherwise this PR gets to big

docs/usage/Context.md
schrodit: no you can only configure the default context that is then populated to all namespaces.
Other contexts should be added manually.
It comparable to the service accounts and the default service account in core k8s.

pkg/landscaper/installations/context.go
schrodit: You're right this is confusing.
@Diaphteiros and me already tried to find a suitable naming here.

External in this case means that the context is defined in an external resource.
So in another CR in the cluster.

Do you have a better idea for the naming?

schrodit:
> I think that makes sense. Could probably also be done in a seperate PR?
yes otherwise this PR gets to big

pkg/landscaper/installations/context.go
schrodit: you mean to the docstring in line 88?

reshnm: yes, exactly

docs/usage/Context.md
schrodit: I think it will be clearer as soon as this PR is also added: https://github.com/gardener/landscaper/pull/335

Then the structure would look like:
```
controllers:
  contexts:
    workers: 5
    config:
      default:
        repositoryContext: ...
```

But we could also rename the `default` key to `defaultContext`. Maybe that will also make it clearer.
WDYT?

reshnm: Ah, ok. No, then i think it is OK to leave it like it is.
I think we just need some good tutorials on how to handle/manage contexts.

reshnm (approved):
/lgtm

Diaphteiros (changes requested):
Some minor things, mostly rephrasing proposals where the documentation could be a bit more clear.

docs/concepts/Glossary.md
```diff
@@ -29,14 +29,15 @@
...
 
-  A context defines the scope in which an Installation runs and all of its data lives.
+  The Landscaper defines 2 different kind of contexts.
+  One context is an actual resource that is referenced by an installation and defines common configuration.
+
+  During the execution of an installation a context defines the scope in which an Installation runs and all of its data lives.
   For every Installation a dedicated context is created and data can only be accessed within the same context.
   Data can be exchanged between contexts via Import and Export declarations.
 
   Since Installations can be nested, the resulting Contexts are nested as well.
```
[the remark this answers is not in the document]
Diaphteiros: (the line markings are wrong, the comment refers to the whole Context explanation)

docs/usage/Context.md
```diff
@@ -0,0 +1,55 @@
...
+
+The context is a configuration resource containing shared configuration for installations.
+This config can contain the repository context, registry pull secrets, or even deployer specific context.
+
+A context can be referenced by installations in the same namespace.
+
+## Basic structure
+
+A context object contains the repository context and an optional list of registry pull secrets.
+These registry pull secrets are references to secrets in the same namespace as the context that contains oci registry access credentials.
```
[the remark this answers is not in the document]
schrodit: I restructured the whole sentence as it was little bit confusing to read.

docs/concepts/Glossary.md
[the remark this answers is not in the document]
schrodit: yes the idea is the resource is capitalized. I updated the paragraph accordingly.

docs/concepts/Glossary.md
```diff
@@ -29,14 +29,15 @@
 
 #### _Context_
 
-  A context defines the scope in which an Installation runs and all of its data lives.
+  The Landscaper defines 2 different kind of contexts.
```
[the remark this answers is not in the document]
schrodit: Done. PTAL

Diaphteiros (approved):
Small nit, lgtm otherwise.

PR #336, opened 2021-10-01
move the subinstallations of landscaper service to seperate files

reshnm:

/area operations
/kind enhancement
/priority 4

This PR moves the subinstallations to individual files, which should increase readability and maintainability.
Added some documentation for blueprint imports and exports.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.landscaper/landscaper-service/blueprint/installation/blueprint.yaml
+++ b/.landscaper/landscaper-service/blueprint/installation/blueprint.yaml
@@ -2,55 +2,68 @@ apiVersion: landscaper.gardener.cloud/v1alpha1
 kind: Blueprint
 
 imports:
+   # The "hostingCluster" specifies the target Kubernetes cluster into which the landscaper service will be installed.
   - name: hostingCluster
     required: true
     targetType: landscaper.gardener.cloud/kubernetes-cluster
 
+  # The "hostingClusterNamespace" is the namespace in the hosting cluster into which the landscaper service is getting installed.
+  # Only one landscaper server per namespace can be installed.
   - name: hostingClusterNamespace
     required: true
     type: data
     schema:
       type: string
 
+  # Specifies whether the "hostingClusterNamespace" is getting deleted before the installation.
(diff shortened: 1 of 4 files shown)


.landscaper/landscaper-service/blueprint/installation/blueprint.yaml
```diff
@@ -2,55 +2,68 @@ apiVersion: landscaper.gardener.cloud/v1alpha1
 kind: Blueprint
 
 imports:
+   # The "hostingCluster" specifies the target Kubernetes cluster into which the landscaper service will be installed.
   - name: hostingCluster
     required: true
     targetType: landscaper.gardener.cloud/kubernetes-cluster
 
+  # The "hostingClusterNamespace" is the namespace in the hosting cluster into which the landscaper service is getting installed.
+  # Only one landscaper server per namespace can be installed.
```
schrodit, line 11: these comments could be also added as schema description.
That would give us the possibility to parse the schema and generate a documentation

```yaml
  - name: hostingClusterNamespace
    required: true
    type: data
    schema:
      description: |
        The "hostingClusterNamespace" is the namespace in the hosting cluster into which the landscaper service is getting installed.
        Only one landscaper server per namespace can be installed.
      type: string
```

or we could maybe think of a dedicated field in the import definition.
```yaml
  - name: hostingClusterNamespace
    required: true
    description: |
        The "hostingClusterNamespace" is the namespace in the hosting cluster into which the landscaper service is getting installed.
        Only one landscaper server per namespace can be installed.
    type: data
    schema:
      type: string
```

schrodit (commented):
one small comment otherwise lgtm

.landscaper/landscaper-service/blueprint/installation/blueprint.yaml
reshnm: 👍 
I moved it to `schema.description`.
Maybe it would be good to have something similar for targets.

schrodit (approved):
/lgtm

PR #338, opened 2021-10-08
do not try to load chart from archive for non-success http status codes

reshnm:

/area helm-deployer
/kind bug
/priority 3

When a non-success status code was received while resolving a helm chart from http access, we have ignored this error and tried to load the chart anyway.
This results in a strange error message:

`unable to load chart from \“https://myrepo/mychart.tgz\“: gzip: invalid header”`

This change introduces a handling for non-success http status codes and returns a better error message.

**Release note**:

```other operator
NONE
```


schrodit (approved):
/lgtm

PR #342, opened 2021-10-14
Blueprint index method

schrodit:

/area landscaper
/kind enhancement
/priority 3

Adds a new index method for the blueprint store that uses the digest of the blueprints has been added and made as default.
It i still possible to use the component descriptor + blueprint name as cache index.

Fixes #

**Release note**:

```feature operator
Adds a new index method for the blueprint store that uses the digest of the blueprints has been added and made as default.
With that change blueprints are updated when the content of a blueprint changes which is useful for development.
```


Diaphteiros (approved):
/lgtm

PR #344, opened 2021-10-18
Fixes a panic for root installations and adds support for targets in the dependency check

schrodit:

/kind bug
/priority 3

Fixes a nil pointer in the validation that occurs for root installations (installations without a parent).
Also added support for targets in the dependency check

**Release note**:

```bugfix user
A bug in the import validation has been fixed that caused the Landscaper to panic when a root Installation is validated.
```
```feature user
The Landscaper now also validates target imports and waits until Installations that have created that target are finished.
```


enrico-kaack-comp (approved):
LGTM

PR #351, opened 2021-10-21
Improve help and error message

Diaphteiros:

/area usability
/kind enhancement
/priority 3

- The error message for imports which were defined in a blueprint but not satisfied in the installation has been improved to better show what the actual problem is.
- The help information on the landscaper-controller command displayed a wrong list of supported deployers (`manifest` was missing). This information is now computed based on the default deployer configurations, so it should always be in sync with the actually supported deployers. Disadvantage of this solution is that the order in which the deployers are listed varies, as they are pulled from a map.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/deployermanagement/config/options.go
+++ b/pkg/deployermanagement/config/options.go
@@ -5,6 +5,7 @@
 package config
 
 import (
+	"fmt"
 	"io/ioutil"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/deployermanagement/config/options.go
```diff
@@ -25,10 +26,16 @@ type Options struct {
 
 // AddFlags adds the flags for the deployer management.
 func (o *Options) AddFlags(fs *flag.FlagSet) {
+	defaultDeployers := make([]string, len(DefaultDeployerConfiguration))
+	i := 0
+	for k := range DefaultDeployerConfiguration {
+		defaultDeployers[i] = k
+		i++
+	}
```
schrodit, line 34:
```suggestion
	defaultDeployers := make([]string, 0)
	for k := range DefaultDeployerConfiguration {
		defaultDeployers = append(defaultDeployers, k)
	}
```

schrodit (changes requested):
just a minor nit.
otherwise lgtm

schrodit (approved):
/lgtm

PR #355, opened 2021-10-28
fix landscaper version in integration test

schrodit:

/kind bug
/priority 3

Fixes a bug in the integration test to now use the real version that is created in the pipeline.

**Release note**:

```other operator
NONE
```


reshnm (approved):
/lgtm

PR #357, opened 2021-10-29
Update helm chart versions before uploading the chart

reshnm:

/area operations
/kind bug
/priority 3

The versions of the helm charts need to be updated before they are uploaded to the OCI registry.
This change moves the version update from step "release" to step "publish-helm-charts".

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #367, opened 2021-11-05
default deployer context

achimweigel:

/area deployers
/kind bug
/priority 3

This PR uses the default repository context of the landscaper to fetch the default deployer (helm, container, manifest). This resolves the problem to fetch their component descriptors from the right location.

Furthermore, it allows to configure an alternative repository context to fetch the component descriptors for the default deployer. This allows to support the use case where the landscaper artefacts are stored in a different repo than the components descriptors and blueprints. 

Fixes #

Manually tested against running landscaper installation.

**Release note**:

```bugfix operator
Component Descriptors of deployers (helm, container, manifest) are fetched from default repository context if no other context is configured.
```


apis/config/v1alpha1/types_landscaper_config.go
```diff
@@ -216,6 +216,11 @@ type DeployerManagementConfiguration struct {
 	Namespace string `json:"namespace"`
 	// Agent contains the landscaper agent configuration.
 	Agent LandscaperAgentConfiguration `json:"agent"`
+	// DeployerRepositoryContext defines the repository context to fetch the component descriptors for the
+	// default deployer (helm, container, manifest). If not set, the default repository context for component
```
schrodit, line 220:
```suggestion
	// default deployers (helm, container, manifest). If not set, the default repository context for component
```

apis/config/v1alpha1/types_landscaper_config.go
```diff
@@ -216,6 +216,11 @@ type DeployerManagementConfiguration struct {
...
+	// descriptors and blueprints is used.
```
schrodit, line 221:
```suggestion
	// default deployer (helm, container, manifest). If not set, the open source repository context for component
	// descriptors and blueprints is used.
```

cmd/landscaper-controller/app/app.go
```diff
@@ -93,6 +95,20 @@ func (o *Options) run(ctx context.Context) error {
 		opts.MetricsBindAddress = fmt.Sprintf(":%d", o.Config.Metrics.Port)
 	}
 
+	if o.Config.DeployerManagement.DeployerRepositoryContext == nil {
+		if o.Config.Controllers.Contexts.Config.Default.RepositoryContext != nil {
+			o.Config.DeployerManagement.DeployerRepositoryContext = o.Config.Controllers.Contexts.Config.Default.RepositoryContext.DeepCopy()
+		} else if o.Config.RepositoryContext != nil {
+			o.Config.DeployerManagement.DeployerRepositoryContext = o.Config.RepositoryContext.DeepCopy()
+		} else {
+			defaultCtx, err := cdv2.NewUnstructured(cdv2.NewOCIRegistryRepository("eu.gcr.io/gardener-project/development", ""))
```
schrodit, line 104: we should make that default in the configuration defaulting here: https://github.com/gardener/landscaper/blob/master/apis/config/v1alpha1/defaults.go#L63

schrodit (changes requested):

cmd/landscaper-controller/app/app.go
achimweigel: I have moved the coding

apis/config/v1alpha1/defaults.go
```diff
@@ -72,6 +74,18 @@ func SetDefaults_LandscaperConfiguration(obj *LandscaperConfiguration) {
...
 		obj.DeployerManagement.Agent.OCI = obj.Registry.OCI
 	}
+
+	if obj.DeployerManagement.DeployerRepositoryContext == nil {
+		if obj.Controllers.Contexts.Config.Default.RepositoryContext != nil {
+			obj.DeployerManagement.DeployerRepositoryContext = obj.Controllers.Contexts.Config.Default.RepositoryContext.DeepCopy()
+		} else if obj.RepositoryContext != nil {
+			obj.DeployerManagement.DeployerRepositoryContext = obj.RepositoryContext.DeepCopy()
+		} else {
+			defaultCtx, _ := cdv2.NewUnstructured(cdv2.NewOCIRegistryRepository("eu.gcr.io/gardener-project/development", ""))
```
reshnm, line 84: Wouldn't it be safer to have the previous default repository context (eu.gcr.io/gardener-project/development) when none is set explicitely?
Will this be backwards compatible?

apis/config/v1alpha1/defaults.go
```diff
@@ -72,6 +74,18 @@ func SetDefaults_LandscaperConfiguration(obj *LandscaperConfiguration) {
 	if obj.DeployerManagement.Agent.OCI == nil {
 		obj.DeployerManagement.Agent.OCI = obj.Registry.OCI
 	}
+
+	if obj.DeployerManagement.DeployerRepositoryContext == nil {
+		if obj.Controllers.Contexts.Config.Default.RepositoryContext != nil {
+			obj.DeployerManagement.DeployerRepositoryContext = obj.Controllers.Contexts.Config.Default.RepositoryContext.DeepCopy()
+		} else if obj.RepositoryContext != nil {
```
schrodit, line 81: not sure if we really want to default to the default repository.
I think this will lead to issues in most cases as the deployers would have to be transported to the default context by the user.

I think we should just default to the open source repository if no context is defined.

apis/config/v1alpha1/defaults.go
```diff
@@ -72,6 +74,18 @@ func SetDefaults_LandscaperConfiguration(obj *LandscaperConfiguration) {
...
+			obj.DeployerManagement.DeployerRepositoryContext = obj.RepositoryContext.DeepCopy()
```
schrodit, line 82: same for this field.

apis/config/v1alpha1/defaults.go
achimweigel: That's right. I have changed it.

apis/config/v1alpha1/defaults.go
achimweigel: I have changed this accordingly

reshnm (approved):
/lgtm

schrodit (approved):

PR #372, opened 2021-11-10
Pass context

schrodit:

/area deployers
/priority 3

The context reference specified in the installation is now passed to the subinstallations, execution and deploy items.
With that it is now possible for the deployers to use namespaced secrets to access oci or other artifacts.

Fixes #

**Release note**:

```feature operator
The context reference is not properly passed to all DeployItems and subinstallations.
With that the deployers are now able to use the secrets defined in the Context to access remote artifacts.
This is especially useful for multi-tenant scenarios where everything should isolated by the namespace.
```


pkg/deployer/lib/controller.go
```diff
@@ -197,6 +200,12 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return returnAndLogReconcileResult(logger, *hookRes), nil
 	}
 
+	lsCtx := &lsv1alpha1.Context{}
+	// todo: check for real repository context. Maybe overwritten by installation.
```
reshnm, line 204: Reading this I have a feeling this could be something that leads to issues on user side by picking the wrong context.
What is currently stopping us from finding the "real" repository context? Is it a matter of missing APIs?

reshnm (commented):
@schrodit Looking at this change I can see that the context is added to a lot of API functions and most of the time it just passed on to the next layer af API.

Wouldn't it be "better" to have something like `GetContextForDeployItem` (optionally with some caching/lazy loading) which could be called just by the places that need to actually read the context?

pkg/deployer/lib/controller.go
schrodit: > What is currently stopping us from finding the "real" repository context? Is it a matter of missing APIs?

The initial idea was that there is default context defined that is used by all installation.
Only for dev or debugging purposes the repository context directly in the installation should be set.
The same idea is then also used in the implemenation with the context object.
The context object is there to namespaced default context and in a production environment the repository should never be set in the installation but always be read from the context.

Coming now to deploy items, which are reconciled by deployers, there is now the problem that if a context is overwritten by the installation this is not reflected int he context object.
And the issue is that the deployer should not know about an installation. The only contract should be a deploy/target and context.

reshnm:
Closed in favour of #392

PR #373, opened 2021-11-10
download component-cli via curl instead of go get

reshnm:

/area operations
/kind bug
/priority 3

Trying to download `component-cli` via ´go get` will fail with

```
go get github.com/gardener/component-cli/cmd/component-cli: create zip: pkg/commands/componentarchive/testdata/01-ca-blob/blobs/sha256:ab894987c426bf8d660826c6fa52a1f351a4c4c094f913862be9c76386bcc32f: malformed file path "pkg/commands/componentarchive/testdata/01-ca-blob/blobs/sha256:ab894987c426bf8d660826c6fa52a1f351a4c4c094f913862be9c76386bcc32f": invalid char ':'
make: *** [cnudie] Error 1
```

It seems that `go get` cant handle certain paths containing the char `:`.
With this PR the script will use curl to download the latest component-cli binary directly.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/hack/generate-cd.sh
+++ b/hack/generate-cd.sh
@@ -19,7 +19,8 @@ fi
 if ! which component-cli 1>/dev/null; then
   echo -n "component-cli is required to generate the component descriptors"
   echo -n "Trying to installing it..."
-  go get github.com/gardener/component-cli/cmd/component-cli
+  curl -L https://github.com/gardener/component-cli/releases/download/$(curl -s https://api.github.com/repos/gardener/component-cli/releases/latest | jq -r '.tag_name')/componentcli-$(go env GOOS)-$(go env GOARCH).gz | gzip -d > $(go env GOPATH)/bin/component-cli
+  chmod +x $(go env GOPATH)/bin/component-cli
 
   if ! which component-cli 1>/dev/null; then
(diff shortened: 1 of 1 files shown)


enrico-kaack-comp:
Even this workaround works in this place, can we fix our package so go get works?

schrodit:
> Even this workaround works in this place, can we fix our package so go get works?

I would not consider this a workaround. This should be the way how the cli is installed.

As for the fix. We can fix it but it would require a change in how the local blob creation in done.
The actual issue resides from testdata for these local blobs, e.g. https://github.com/gardener/component-cli/tree/main/pkg/commands/componentarchive/testdata/01-ca-blob/blobs .

achimweigel:
I just tried to replaced the ":" by a "-" in the blob file name and the test suite componentarchive_suite_test.go still worked. Not sure but perhaps this is a quite simple solution.

achimweigel:
> I just tried to replaced the ":" by a "-" in the blob file name and the test suite componentarchive_suite_test.go still worked. Not sure but perhaps this is a quite simple solution.

Geht doch nicht, da der Blob noch in einem anderen Test verwendet wird :-(

schrodit (approved):

PR #375, opened 2021-11-11
set default repo context of cd imports

achimweigel:

/area usability
/kind enhancement
/priority 3

This PR adds the default repo context to component descriptor imports if no other context is set. This reduces the complexity to define installations and prevents landscaper restarts due to nil pointer dereferencing.

Fixes #

**Release note**:

```other operator
Default context for component descriptor imports
```


Diaphteiros (approved):
/lgtm

PR #381, opened 2021-11-23
Fix helm chart templating for targetSelector

robertgraeff:

/area deployers
/kind bug
/priority 3

The targetSelectors of the deployers are configured in the helm chart values at `.deployer.targetSelector`.  This configuration gets currently lost, because the templating of the charts uses the wrong path `.Values.targetSelector` instead of the correct path `.Values.deployer.targetSelector`. The PR fixes this.

Fixes #

**Release note**:

```bugfix operator
NONE
```


achimweigel (approved):
/lgtm

PR #387, opened 2021-12-03
add namespace read permissions as needed by the context controller

reshnm:

/area operations
/kind bug
/priority 3

The new context controller needs read access for namespaces (https://github.com/gardener/landscaper/blob/master/pkg/landscaper/controllers/context/add.go#L39).
This permission is missing from the landscaper controller clusterrole.

```
E1203 11:31:25.756253       1 reflector.go:138] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:255: Failed to watch *v1.Namespace: failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:ls-system:landscaper-controller" cannot list resource "namespaces" in API group "" at the cluster scope
```

This symptom seems to only affect the landscaper service use case as it is otherwise masked by the landscaper agent clusterrole which has permissions to basically all resources.

**Release note**:

```bugfix user
Add namespace read permissions to the landscaper controller clusterrole in the landscper rbac chart.
```

### DIFF
--- a/charts/landscaper/charts/rbac/templates/clusterrole-controller.yaml
+++ b/charts/landscaper/charts/rbac/templates/clusterrole-controller.yaml
@@ -52,6 +52,14 @@ rules:
       - update
       - delete
   - apiGroups:
+      - ""
+    resources:
+      - "namespaces"
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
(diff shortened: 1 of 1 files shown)


Diaphteiros (approved):
/lgtm

PR #393, opened 2022-01-12
export ReferenceResolver type and detect cyclic dependencies

Diaphteiros:

/area robustness
/kind enhancement
/kind api-change
/priority 3

This PR does two things:
1. The `ReferenceResolver` type is now public. This allows for better coding in the landscaper CLI tooling.
2. Cyclic references will now be detected instead of causing endless recursion.

Prerequisite for the landscaper CLI fix.

**Release note**:

```bugfix operator
Cyclic references in jsonschema files are now detected and cause a corresponding error instead of getting the landscaper stuck in endless recursion.
```

### DIFF
--- a/pkg/landscaper/jsonschema/reference.go
+++ b/pkg/landscaper/jsonschema/reference.go
@@ -40,17 +40,17 @@ type ReferenceContext struct {
 	ComponentResolver ctf.ComponentResolver
 }
 
-type referenceResolver struct {
+type ReferenceResolver struct {
 	*ReferenceContext
 }
 
-func NewReferenceResolver(refCtx *ReferenceContext) *referenceResolver {
-	return &referenceResolver{refCtx}
+func NewReferenceResolver(refCtx *ReferenceContext) *ReferenceResolver {
+	return &ReferenceResolver{refCtx}
 }
 
 // Resolve walks through the given json schema and recursively resolves all references which use one of
 // the "local", "blueprint", or "cd" schemes.
-func (rr *referenceResolver) Resolve(schemaBytes []byte) (interface{}, error) {
+func (rr *ReferenceResolver) Resolve(schemaBytes []byte) (interface{}, error) {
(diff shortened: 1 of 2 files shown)


pkg/landscaper/jsonschema/reference.go
```diff
@@ -132,18 +135,26 @@ func (rr *referenceResolver) resolveList(data []interface{}, currentPath *field.
 }
 
 // resolveReference resolves a reference
-func (rr *referenceResolver) resolveReference(s string, currentPath *field.Path) (interface{}, error) {
+func (rr *ReferenceResolver) resolveReference(s string, currentPath *field.Path, alreadyResolved stringSet) (interface{}, error) {
 	uri, err := url.Parse(s)
 	if err != nil {
 		return nil, fmt.Errorf("invalid url: %w", err)
 	}
+	if rr == nil || rr.ReferenceContext == nil {
```
reshnm, line 143: I don't understand why this is checked here.
`rr` is the this pointer and is used in a non-exported "private" member function of `ReferenceResolver`.
I don't see why it can be nil at this point.

For the `rr.ReferenceContext`, this field is getting initialized in `NewReferenceResolver` if it is passed as nil by the user.
It can only be nil when someone explicitely sets the pointer to nil.
If it needs to be checked for nil, doesn't it make more sense to do it in function `Resolve`?

reshnm (changes requested):

pkg/landscaper/jsonschema/reference.go
Diaphteiros: I did this to avoid some problems with the tests: 
Before, rr.ReferenceContext being nil was no problem, but after I added the cycle check, it didn't work anymore. Therefore I added this check and I forgot to remove it, after I added the defaulting in the constructor. I guess we don't need this check anymore.

reshnm (approved):
/lgtm

PR #394, opened 2022-01-13
Render blueprint

robertgraeff:

/area dev-productivity
/kind enhancement
/priority 3

This pull request is a preparation for a follow-up PR of the Landscaper CLI. The goal is to improve the  `landscaper blueprint render...` command. 

- The current PR enhances function `RenderBlueprint`. It is now possible to pass a component descriptor as object. Previously, a component descriptor could only be passed via a path to a component descriptor file.
- The current PR enhances function `RenderBlueprintDeployItems`. It is now possible to pass a component resolver to load resources specified in the component descriptor.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/utils/landscaper/blueprint.go
+++ b/pkg/utils/landscaper/blueprint.go
@@ -5,6 +5,7 @@
 package landscaper
 
 import (
+	"context"
 	"fmt"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


Diaphteiros (approved):
/lgtm

PR #395, opened 2022-01-19
Prepare gardener

Diaphteiros:

/area robustness
/kind enhancement
/priority 3

Currently, the landscaper checks whether all dependents are in a final phase before starting to reconcile an installation. However, in theory this could lead to a race condition when multiple installations are updated simultaneously - if a depending installation performs this check after its dependents have been updated, but before the landscaper has put them into the pending phase, it will be reconciled as everything seems alright. This is not intended.

With this PR, the check now also verifies that `Status.ObservedGeneration == Generation`, which should prevent the mentioned case.

**Release note**:

```other operator
Before starting to reconcile an installation, landscaper now checks whether any installation it depends on has had an update which has not yet been processed and blocks the reconciliation if this is the case. This is identified by comparing the installations' `generation` to the `observedGeneration` in their status fields.
```

### DIFF
--- a/pkg/landscaper/installations/imports/helper.go
+++ b/pkg/landscaper/installations/imports/helper.go
@@ -86,6 +86,11 @@ func CheckCompletedSiblingDependents(ctx context.Context,
 			return false, false, nil
 		}
 
+		if inst.Generation != inst.Status.ObservedGeneration {
+			log.V(3).Info("dependent installation completed but not up-to-date", "inst", sourceRef.NamespacedName().String())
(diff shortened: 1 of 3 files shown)


Diaphteiros:
/hold
Don't merge yet, we need to investigate whether this change has unwanted/unexpected side effects.

Diaphteiros:
cc @achimweigel @robertgraeff

achimweigel (approved):
/lgtm

PR #406, opened 2022-02-01
Fix reconcile logic

Diaphteiros:

/area robustness
/kind bug
/kind api-change
/priority 3

**Release note**:

```bugfix operator
Fixes a bug which could lead to landscaper and a deployer infinitely reconciling a deployitem without it ever reaching a final state.
```
```other developer
The status of an execution now contains a list referencing deployitems which stores the generation the execution had when it last applied the respective deployitem. This is used to detect changes in the execution which have not been propagated to the deployitems. The `ObservedGeneration` in the previously existing list of deployitem references now refers to the generation which the deployitem had after it had last been applied by the execution and is used to detect manual changes to deployitems.
```


reshnm (approved):
/lgtm

Issue #408, opened 2022-02-02
Manual changes to an execution are not reverted


Diaphteiros:
**How to categorize this issue?**

/area robustness
/kind bug
/priority 3

**What happened**:
I created an installation, which has a subinstallation, which creates an execution with one deployitem. Then I modified the execution manually and made a change to the deployitem template. Landscaper did not revert the change done to the execution. This could only be enforced by using a force-reconcile annotation on the subinstallation.

**What you expected to happen**:
I would expect that landscaper automatically recognizes changes to executions and reverts them.
Even if it doesn't, this should at least happen with a reconcile Annotation on the root installation, but not only did that one not help, but even a force-reconcile annotation on the root installation did not enforce recreation of the execution.
A force-reconcile annotation on the subinstallation helped, but a user should not have to search for the correct subinstallation, after he accidentally messed up an execution.

**Anything else we need to know?**:
Some proposals how we could solve this:
1. Landscaper automatically recognizes changes to executions and reverts them, as it does with deployitems. This would be the 'nicest' solution, but it might require additional status fields and logic to detect changes.
2. Landscaper does not revert changed executions on its own, but force-reconcile annotations on installations are at least propagated to all of their subinstallations, so that a force-reconcile annotation on the root installation will re-create the execution and overwrite any changes.

Diaphteiros:
cc @achimweigel @robertgraeff @reshnm

In-Ko:
Moved to new backlog project, therefore closing this here.

PR #412, opened 2022-02-08
Ignore annotation and operation annotation improvements

### LINKED ISSUE #399: Support "StopReconcile" Annotation (was "Explicit reconcile order for root installations")

/area deployers
/kind enhancement
/priority 3

Assume you have several to root installations, e.g. I1, I2, I3 where I3 import data from I2 and I2 from I1. Now you want to upgrade the version of the referenced Component Descriptor / Blueprint in I1 and I3. Thereby it is critical, that the upgrade of I1 is executed before I3. With the current solution this is not possible and the client must coordinate the upgrade order for the installations. 

To get more control of the upgrade/reconcile order a mechanism is needed to stop the reconcile of a set of root installations, upgrade some of them and then start the reconcile of all these installations in a predefined order. 

This feature was requested by the Gardener team for the deployment of Gardener with Landscaper. 
 
**Why is this needed**:

Diaphteiros:

/area usability
/kind enhancement
/priority 2

Adds an 'ignore' annotation to disable landscaper operations on specific resources.
Also slightly modifies the effects of operation annotations - especially force-reconcile - to be more consistent.
Adds documentation for the annotations.

Part of #408 
Fixes #399 

**Release note**:

```feature user
It is now possible to annotate Installations, Executions, and DeployItems with `landscaper.gardener.cloud/ignore: true`. This will prevent reconciliation of the respective resource. Please note that this blocks reconciliation of depending resources and the annotation is neither set nor removed by Landscaper itself. See the new [Annotations documentation](docs/usage/Annotations.md) for details.
```
```bugfix user
The `landscaper.gardener.cloud/operation: force-reconcile` annotation on an Installation is now propagated to all subinstallations of that installation. This ensures that not only the direct Executions of that Installation are reconciled, but also the transitive ones.
```

### DIFF
--- a/docs/usage/Annotations.md
+++ b/docs/usage/Annotations.md
@@ -21,7 +21,7 @@ The operation annotation is removed during the reconciliation.
 **force-reconcile**
 This enforces a redeployment of executions and subinstallations. The checks, whether any of them is still progressing or the installation's imports are outdated, are skipped.
 In order to fix potentially broken executions of subinstallations, the force-reconcile annotation will be propagated to the subinstallations.
-The operation annotation is removed during the force-reconciliation.
+The operation annotation is removed during the reconciliation.
 
 **abort**
 If the abort operation annotation is set, a reconcile will be stopped before checking whether the installation needs to be updated.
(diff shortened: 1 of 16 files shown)


docs/usage/Annotations.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Ignore Annotation
+
+**Annotation:** `landscaper.gardener.cloud/ignore`
+**Accepted values:**
+  - `true`
+
+#### Effect on Installations/Executions/DeployItems
+
+The effect of this annotation is the same for all landscaper resources: the respective resource will not be reconciled by the landscaper, even if its spec changed or the operation annotation says otherwise. Only resources in a final phase are affected, to interrupt a running installation/execution/deployitem, the `landscaper.gardener.cloud/operation=abort` annotation has to be used.
```
reshnm, line 70: "Only resources in a final phase are affected [...]"

I think I understand what is meant here. However I think in the way it is written here, it is not entirely correct.
Resources that have no phase yet are also affected by the annotation, aren't they? Means, when someone creates a resource with the ignore annotation it will be blocked and the resource has no phasse assigned yet.

Shouldn't it be like this?

The annotation will not affect resources that are in a non-final phase, to interrupt a running (`Init`, `Progressing`or `Deleting`) installation/execution/deployitem, the `landscaper.gardener.cloud/operation=abort` annotation has to be used.

reshnm (approved):
/lgtm

PR #414, opened 2022-02-14
documentation refactoring and aligning code with concept

Diaphteiros:

/area usability
/area documentation
/kind enhancement
/priority 3

Most importantly, this PR contains a rewrite of the central documentation.

Furthermore, some functionalities are added/changed:

**1.** Blueprints now contain an `importExecutions` field. This allows
  **a)** using templating to validate the imports (in addition to jsonschema) and generate error messages for wrong inputs
  **b)** hard-coding structural changes for imports (e.g. combining two single imports into one bigger struct on blueprint level)

**2.** The value of `targetType` fields in blueprints is now prefixed with `landscaper.gardener.cloud/`, if it does not contain a `/`. The advantage is that target import declarations in blueprints are now shorter (only need `kubernetes-cluster` instead of `landscaper.gardener.cloud/kubernetes-cluster`) for the built-in types. :warning: This change is not fully backward compatible.

**3.** The names for dataobject and target exports were previously hashed. This is necessary for subinstallations to ensure the correct scope, but it was also done for root installations. This is has partially changed now: if the context name is empty (as it is the case for root installations), the name will only be hashed if it is incompatible with the k8s naming scheme. There are several consequences:
  **a)** The `#` notation for dataobject/target imports to use non-hashed names is still valid, but not required anymore.
  **b)** Root level installation exports can now overwrite manually created dataobjects/targets (technically, this was possible before, but because of the hashed names for exports, naming conflicts were unlikely). This behaviour is desired, though - the core idea of landscaper is that it doesn't care where an import comes from and whether it is an export or created by a user and having to use `#` for user-made objects contradicts this idea. It should be possible to have an installation export replace a previously manually created dataobject/target.
  **c)** Before, root installation exports (or imports) could define `dataRef`s with more-or-less arbitrary names, as the name was hashed before creating the dataobject resource in the cluster. If the names now simply aren't hashed anymore, this will break all installations were the `dataRef` field contains a name that cannot be used for k8s resources (e.g. because it has a capital letter in it). Since this would break a lot of installations, including our tutorial ones, names for root installation datarefs will still be hashed if they are incompatible with the k8s naming scheme otherwise. This applies to target exports and imports too.

**4.** The `target` field in deployitem templates in the blueprint previously took an object reference (`name` & `namespace`). To use an imported target for this, one would have to extract this information from the k8s metadata (`.imports.mytarget.metadata.name`). While a `name` can still be specified in the deployitem template's `target` field, it now also allows to specify an `import` instead (optionally with `index` for targetList imports) to directly reference an imported target by name. Landscaper then converts this import reference into an object reference when the Execution object is created. The `name` field is deprecated and will be removed in the future. It will then not be possible anymore to use targets in deployitems which have not been imported by the blueprint/installation, but we don't really want that to happen anyway.

**5.** The templating has been slightly refactored and the bindings for template executions have been made more consistent. Additionally, an `include` function, similarly to how helm does it, has been added to the `go template` templater.

**Release note**:

```feature user
A new field `importExecutions` has been added to the blueprint. It allows restructuring import values, similar to `importDataMappings`, but on blueprint level.
```
```feature user
Targets can (and should) now be referenced by their import name (and optionally index, in case of targetlist imports) in the deployitem templates in the blueprint. The old way of referencing a target by name and namespace is still supported for backward compatibility, but this will be removed in the future.
```
```feature user
An `include` function has been added to the go templater. It behaves identical to the synonymous function in helm templates.
```
```breaking user
Target types referenced in target import definitions in a blueprint will now be prefixed with `landscaper.gardener.cloud/` if they don't contain a `/`. If all target types you used contained a `/`, you don't have to do anything. Otherwise, please adapt your target types to contain a `/`, or you will get errors during import validation.
```
```feature user
Previously, the names of exported dataobjects and targets were hashed and the name had to be prefixed with a `#` in the installation imports if a manually created (and thus un-hashed) dataobject/target was to be imported. This is now not anymore the case _for imports and exports of root installations_, except for import/export names which are not valid names for k8s resources.
 The `#` notation is still supported for backward compatibility, but it has no effect anymore.

Names of imports/exports of nested installations are still hashed to avoid conflicts.

:warning: Beware that root-level installation exports can now overwrite manually created dataobjects/targets, if the export name matches the name of the resource (technically, this was possible before too, but unlikely due to the hashing of the export name).
```
```other user
Major parts of the documentation have been refactored and updated to more precisely describe relevant fields and allowed values.
```


docs/usage/AccessingBlueprints.md
```diff
@@ -0,0 +1,74 @@
+# Remote Access
+
+Blueprints are referenced in installations or installation templates via the component descriptors access.
+
+Basically blueprints are a filesystem, therefore, any blob store could be supported.
```
achimweigel, line 5: I would remove this sentence - I was confused by it.

docs/usage/Targets.md
```diff
@@ -0,0 +1,7 @@
+# Targets
+
+Targets are a specific type of import and contains additional information that is interpreted by deployers.
```
achimweigel, line 3:
```suggestion
Targets are a specific type of import and contain additional information that is interpreted by deployers.
```

docs/usage/Templating.md
```diff
@@ -0,0 +1,245 @@
...
+    file: "data/template"
+  ```
+- Template file
+  ```
+  deployitems:
+    - name: my-second-deploy-item
+      type: landscaper.gardener.cloud/mock
+      config:
+  {{ include "data/config" . | indent 6 }}
+  ```
```
achimweigel, line 82: Why do we only describe the access to the blueprint file system. We have other access possibilities in the template like the imports, component descriptor - or do I misunderstand here something?

achimweigel (changes requested):

docs/usage/Templating.md
achimweigel: perhaps just a link would be sufficient

docs/usage/AccessingBlueprints.md
Diaphteiros: I rephrased it to be more clear. WDYT?

docs/usage/Templating.md
Diaphteiros: I added a paragraph with a link pointing to the blueprint rendering docs.

achimweigel: 👍

docs/usage/AccessingBlueprints.md
achimweigel: For me the part with the filesystem was confusing at the beginning. What about adding (not sure if this is technically ok): Basically, blueprints are a filesystem, i.e. a folder containing a file named "blueprint.yaml" with the blueprint manifest and additional files and subfolders containing whatever needed, like configuration data etc.

docs/concepts/Glossary.md
```diff
@@ -2,28 +2,30 @@
...
+  other blueprints.
+  They contain an orchestration pattern for nested [_Installations_](#installation).
 
-  Their pratical use is to install mutliple components that depend on each other.
+  Their practical use is to install multiple components that depend on each other.
 
 #### _Blueprint_
 
-  Blueprints contain actual instructions and steps on how to install a software component and what is needed to perform these actions.
+  Blueprints contain actual instructions and steps on how to install a software component form of [_DeployItems_](#deployitem).
```
achimweigel, line 13:
```suggestion
  Blueprints contain actual instructions and steps on how to install a software component in form of [_DeployItems_](#deployitem).
```

docs/concepts/Glossary.md
```diff
@@ -2,28 +2,30 @@
...
+  They contain an orchestration pattern for nested [_Installations_](#installation).
 
-  Their pratical use is to install mutliple components that depend on each other.
+  Their practical use is to install multiple components that depend on each other.
 
 #### _Blueprint_
 
-  Blueprints contain actual instructions and steps on how to install a software component and what is needed to perform these actions.
+  Blueprints contain actual rules to describe target state for described software
+  installations in form of [_DeployItems_](#deployitem).
```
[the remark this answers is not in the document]
achimweigel: Very good - this will become my template for an explanation what a blueprint is.

achimweigel (approved):

robertgraeff (approved):

PR #430, opened 2022-03-02
rework component version overwrites

Diaphteiros:

/area usability
/kind enhancement
/priority 3

The 'old' component overwrites were specified cluster-wide, opposed to all other configurations where Landscaper assumes a namespace to be an independent tenant.

To solve this in a backward compatible manner, the old implementation has not been changed. It still works as before, but is deprecated now. A new namespace-scoped resource called `ComponentVersionOverwrites` has been introduced, which looks similar to the old format, although there are some differences in behaviour.

Not finished yet, open TODOs:
- [x] write unit tests for the overwrite logic
- [x] documentation
- [x] we need further (integration?) tests to ensure that also the references of CDs are overwritten

**Release note**:

```breaking user
The `ComponentOverwrites` custom resource has been removed and is replaced by the new `ComponentVersionOverwrites` custom resource, which is namespace-scoped. If you upgrade from an existing Landscaper installation, you have to migrate all `ComponentOverwrites` resources to the new format and delete the `ComponentOverwrites` CRD afterwards. See the [documentation](docs/usage/ComponentOverwrites.md) for further information.
```


Diaphteiros:
cc @mandelsoft @reshnm

apis/core/v1alpha1/types_component_version_overwrite.go
```diff
@@ -0,0 +1,73 @@
...
+	Scope:   lsschema.NamespaceScoped,
+	Storage: true,
+	Served:  true,
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ComponentVersionOverwrites are resources that can hold any kind json or yaml data.
```
reshnm, line 43: This comment seems like a copy/paste error?

docs/usage/ComponentOverwrites.md
```diff
@@ -1,11 +1,65 @@
...
+- **`componentName`** *string* (optional)
+- **`version`** *string* (optional)
+- **`repositoryContext`** *[structure](./RepositoryContext.md)* (optional)
+
+If used to match a replacement source, unspecified attributes match any value and specified attributes describe a concrete match.
+If used to describe a replacement target, specified attributes replace the respective attribute in the source, while unspecified attributes are left unchanged.
+
+A list of overwrite specifications is evaluated in the given order. If a source specification matches and none of the given substitution attributes have already been substituted by an earlier match, the substitution is executed.
+
+This means that more specific overwrite specifications should be placed before more generic ones and if a substitution already replaced an attribute, more general substitutions later on, affecting the same field, will be ignored.
```
reshnm, line 27: Could you maybe add examples for this?
I think I get what you are saying, however, having examples would help alot.

docs/usage/ComponentOverwrites.md
```diff
@@ -1,11 +1,65 @@
...
+    repositoryContext:
+      type: ociRegistry
+      baseUrl: "example.org"
+    componentName: ""
+    version: ""
+```
+
+## ComponentOverwrites
+
+A component overwrite object is a cluster-wide specification of component version overwrites. There might be any number of such overwrite objects. The described overwrite specifications are concatenated in the inversed order of the creation timestamps of the resources. This means that the most recent overwrite specifications will be executed before the older ones.
```
reshnm, line 60: Sry to say this, but this sounds like a nightmare to use and to debug :-(

reshnm (changes requested):
To me the whole concept of the component overwrites sounds much too complex to use.
I would like to have a seperate talk about what we want to achieve whith this concept and then find an easy-to-use solution.

apis/core/types_component_version_overwrite.go
```diff
@@ -0,0 +1,56 @@
+// SPDX-FileCopyrightText: 2020 SAP SE or an SAP affiliate company and Gardener contributors.
```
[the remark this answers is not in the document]
reshnm: nit

reshnm (approved):
/lgtm

PR #436, opened 2022-03-07
reference context takes an explicit repository context used for reference resolving

reshnm:

/area robustness
/kind enhancement
/priority 3

Previously the JSON schema reference resolver was only using the effective repository context of the component descriptor, ignoring the explicit setting of the repository context via Landscaper Context, Installation or ComponentOverwrite.
This PR changes that the reference resolver is now using the correct repository context as configured for an Installation.

Fixes #

**Release note**:

```other operator
NONE
```


Diaphteiros (approved):
/lgtm

PR #438, opened 2022-03-08
fix regression with annotations on the same level as references in JSON schemas

reshnm:

/area usability
/kind regression
/priority 3

JSON Schema allows to add annotations to any location within a schema. https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.7.7
There is a regression with the recent changes to reference resolving in JSON Schemas, which is fixed by this PR.

Fixes #

**Release note**:

```bugfix user
Fixed regression that prevented the usage of annotation on the same level as references ($ref) in JSON schemas.
```


Diaphteiros (approved):
/lgtm

achimweigel:
I just started the tests to see, if these are also failing in this PR. We have problems with the tests for the helm deployer.

PR #439, opened 2022-03-09
Rework installation reconcile logic

Diaphteiros:

/area robustness
/kind enhancement
/priority 3

The reconcile flow for installations was hard to understand before. Especially, whether an installation should be reconciled or has to wait. The reason for that was - at least partly - that the checks were distributed over multiple, partly nested, method calls and checking was mixed with actual action (e.g. loading the imports).
This refactoring aims at improving the readability of the code for the reconciliation flow. To achieve this, a new construct has been added which is instantiated during the reconciliation and used to check the different requirements. By storing any information which has been computed or fetched from the cluster, duplicate computation and redundant api calls are avoided. This also allows to better separate the code for checking the requirements and for fetching the necessary information to do so.

**Release note**:

```other developer
The reconciliation logic for installations has been refactored. There should only be minor changes in the Landscaper behavior, if any.
```
```other user
The change from https://github.com/gardener/landscaper/pull/412 that the force-reconcile operation annotation is propagated to nested installations has been reverted, as it causes the nested installations to be applied in arbitrary order, ignoring their dependencies.
```


pkg/landscaper/dataobjects/target.go
```diff
@@ -116,3 +121,36 @@ func (t Target) Apply(raw *lsv1alpha1.Target) error {
...
+
+func (t *Target) GetInClusterObject() client.Object {
+	return t.Raw
+}
+func (t *Target) GetInClusterObjects() []client.Object {
+	return nil
+}
+
+func (t *Target) ComputeConfigGeneration() string {
+	return strconv.Itoa(int(t.GetInClusterObject().GetGeneration()))
```
reshnm, line 143:
```suggestion
	return strconv.FormatInt(t.GetInClusterObject().GetGeneration(), 10)
```

pkg/landscaper/installations/reconcilehelper/reconcilehelper.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+	return updateRequired, nil
+}
+
+// UpdateAllowed returns an error if the installation cannot be updated due to dependencies or unfulfilled imports.
+func (rh *ReconcileHelper) UpdateAllowed() error {
+	err := rh.InstallationsDependingOnReady()
+	if err != nil {
+		// at least one of the installations the current one depends on is not succeeded or has pending changes
+		return installations.NewNotCompletedDependentsErrorf(err, "not all installations which is depended on are ready: %s", err.Error())
```
reshnm, line 97:
```suggestion
		return installations.NewNotCompletedDependentsErrorf(err, "not all installations which are depended on are ready: %s", err.Error())
```

pkg/landscaper/installations/reconcilehelper/reconcilehelper.go
```diff
@@ -0,0 +1,564 @@
...
+					return returnAndSetCondition(false), nil
+				}
+			}
+		}
+	}
+
+	return returnAndSetCondition(true), nil
+}
+
+// InstallationsDependingOnReady returns true if all installations the current one depends on are
```
reshnm, line 188:
```suggestion
// InstallationsDependingOnReady returns nil if all installations the current one depends on are
```

pkg/landscaper/installations/reconcilehelper/reconcilehelper.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	return "", fmt.Errorf("owner reference %q refers to an installation which is neither the parent nor a sibling", ref.NamespacedName().String())
+}
+
+func ownerReferenceIsInstallation(owner *metav1.OwnerReference) bool {
+	return owner != nil && owner.Kind == "Installation"
+}
+
+// getConfigGenerationsFromImportStatus the config generation(s) for the given import
```
reshnm, line 431:
```suggestion
// getConfigGenerationsFromImportStatus gets the config generation(s) for the given import
```

pkg/landscaper/installations/reconcilehelper/reconcilehelper.go
```diff
@@ -0,0 +1,564 @@
...
+		// there is no config generation for component descriptor imports
+	default:
+		return "", nil, fmt.Errorf("unknown import type %q", imp.GetImportName())
+	}
+	// errors while fetching the import status are ignored
+	// as an non-existing import status most probably means that it belongs to a new import which hasn't been imported before
+	return configGen, configGens, nil
+}
+
+func (rh *ReconcileHelper) checkStateForImport(impPath *field.Path, imp dataobjects.Imported) error {
```
reshnm, line 471: Missing docstring?

reshnm (changes requested):
Just some minor nits.
LGTM otherwise.

docs/technical/deployer_contract.md
```diff
@@ -117,15 +117,14 @@ As explained above, even if the type is correct, the deployer might still not be
 There are two important annotations that need to be handled by the deployer: 
 The operation annotation `landscaper.gardener.cloud/operation` indicates that either a human operator or the landscaper wants a specific operation to be fulfilled on this deploy item. The value of the annotation specifies the expected operation:
 - `reconcile`: This deploy item needs to be reconciled. The deployer has to remove this annotation. In addition, it should set the deploy item's phase to `Init` to show the beginning of a new reconciliation and avoid loss of information in case the deployer dies immediately after removing the annotation. In the status, `lastReconcileTime` has to be set to the current timestamp (this value is used to recognize when a deployer is 'stuck' processing a deploy item).
+- `force-reconcile`: This is similar to `reconcile`, but the reconciliation should be 'enforced'. What exactly that means is up to the deployer. It requires the same handling as the reconcile operation annotation.
```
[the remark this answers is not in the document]
Diaphteiros: I can rephrase this a bit, but there is still some deployer specific logic in there. For example, depending on what exactly the deployer does, 'reconcile' could mean 'update all resources to their desired state', while 'force-reconcile' could be implemented as 'delete all resources and deploy them again'. 

The idea remains of course the same, that 'force-reconcile' is basically a more radical 'reconcile, but whoever implements the controller decides how exactly that looks like.

I will try to rephrase this in a way that captures the intention better.

docs/technical/reconciliation_flow.md
```diff
@@ -0,0 +1,273 @@
...
+- ✔️
+  - The check mark stands for 'action' and means that in this step, the most important part of the reconciliation flow is performed. The exact meaning depends on the type of resource which is being reconciled. For example, during the installation reconcile, the most important action would be the creation/update/deletion of the nested installations and executions. This is what the reconciliation is actually supposed to do, everything else around it is just there to ensure that this happens under the correct conditions.
+<!--- 🚫
+  - This symbol represents a potential error. If it is attached to a step together with some description, it means that this step could result in an error and the description gives further information on how the error looks like. As stated above, please note that this is only used to mark *some* of the potential errors, not *all* of them. Even steps without this symbol can abort the reconciliation with an error. For this reason, errors are not marked as potential outcome for checks (in combination with ❔) or on top-level steps which are split up in smaller ones.
+-->
```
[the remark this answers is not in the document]
Diaphteiros: I planned to add this sometime later, because the information which error message can be caused from which steps could be really helpful for debugging. But I doubt that I will do that in the near future, so I can remove the paragraph as well.

reshnm (approved):
/lgtm

PR #450, opened 2022-03-23
Integrationtest version check

robertgraeff:

/area testing
/kind bug
/priority 3

Fixes the integration test and the failing release job.

At the time when the integration test runs during a release job, the change which increases the version numbers in the `Chart.yaml` files of the Landscaper helm charts has not yet been committed. Therefore, the new numbers are not set if the test machinery clones the Landscaper repository and runs the tests based on the cloned repository. The check, that expects the new version number, must therefore fail. This pull request removes the check. 

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #451, opened 2022-03-24
add installation simulator

reshnm:

/area usability
/kind enhancement
/priority 3

This PR refactores the blueprint renderer util, to make the usage of its API more clear.
The previous version had baked many different uses-cases baked into a single function, which made it very hard to understand.
On to of this refactored blueprint renderer, there is now a installation simulator.
The installation simulator takes any blueprint and a set of imports to simulate the installation run with all its deploy items and sub-installations.
The deploy items can be "simulated" by a list of user provided deploy item export templates.
These are go templates that render a map of exports which are applied to a deploy items exports.

Fixes #

**Release note**:

```feature user
Add installation simulator, which allows to simulate complete landscaper installation flows including its deploy items and sub-installations.
```


pkg/landscaper/blueprints/resolve.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // Resolve returns a blueprint from a given reference.
 // If no fs is given, a temporary filesystem will be created.
-func Resolve(ctx context.Context, resolver ctf.ComponentResolver, cdRef *lsv1alpha1.ComponentDescriptorReference, bpDef lsv1alpha1.BlueprintDefinition) (*Blueprint, error) {
+func Resolve(ctx context.Context, resolver ctf.ComponentResolver, cdRef *lsv1alpha1.ComponentDescriptorReference, bpDef lsv1alpha1.BlueprintDefinition, repositoryContext *cdv2.UnstructuredTypedObject) (*Blueprint, error) {
```
achimweigel, line 30: With this additional parameter, the coding becomes even more complicated especially if you only want to see on its interface. Why not providing just the cdRef with the right repo context. Then the calling method is responsible for providing the right data and this is not mixed into this already very complicated method.

pkg/landscaper/controllers/installations/controller.go
```diff
@@ -174,7 +174,7 @@ func (c *Controller) initPrerequisites(ctx context.Context, inst *lsv1alpha1.Ins
 			currOp, "SetupRegistries", err.Error())
 	}
 
-	intBlueprint, err := blueprints.Resolve(ctx, op.ComponentsRegistry(), lsCtx.External.ComponentDescriptorRef(), inst.Spec.Blueprint)
+	intBlueprint, err := blueprints.Resolve(ctx, op.ComponentsRegistry(), lsCtx.External.ComponentDescriptorRef(), inst.Spec.Blueprint, lsCtx.External.RepositoryContext)
```
achimweigel, line 177: Why do we need to add lsCtx.External.RepositoryContext? This should be also part of lsCtx.External.ComponentDescriptorRef()?

pkg/landscaper/installations/helper.go
```diff
@@ -116,7 +116,7 @@ func CreateInternalInstallationWithContext(ctx context.Context,
 	if err != nil {
 		return nil, err
 	}
-	blue, err := blueprints.Resolve(ctx, compResolver, lsCtx.ComponentDescriptorRef(), inst.Spec.Blueprint)
+	blue, err := blueprints.Resolve(ctx, compResolver, lsCtx.ComponentDescriptorRef(), inst.Spec.Blueprint, lsCtx.RepositoryContext)
```
achimweigel, line 119: same as above

pkg/utils/landscaper/installation_simulator.go
```diff
@@ -0,0 +1,677 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var encoded map[string]interface{}
+	err = yaml.Unmarshal(raw, &encoded)
+	if err != nil {
+		return nil, err
+	}
+	return encoded, nil
+}
```
achimweigel, line 677: I think such a simulator is a good idea. The only concern I have is, that even when the coding of the simulator uses the coding of the core landscaper, there is so much additional coding. Don't you have the fear that the simulator does not compute the same as the core stuff?

achimweigel (changes requested):

pkg/utils/landscaper/installation_simulator.go
reshnm: I do agree with you and I share the same concern. The simulator and the landscaper core coding may diverge.
However I currently see no other way to implement such functionality when the landscaper core coding relies directly on the K8S API (which is of course understandable).

In the tests for the installation simulator I tried to use all the features that the landscaper currently provides, so that we hopefully see when things diverge.

achimweigel (approved):
/lgtm

PR #453, opened 2022-03-24
Delete without uninstall

robertgraeff:

/area robustness
/kind enhancement
/priority 3

With this PR it is possible to delete an installation and the corresponding execution and deploy items without uninstalling resources from the target cluster.

Related to "Landscaper resources cannot be deleted if target is not reachable (73)"

**Release note**:

```other operator
- Support for an annotation to delete an installation without uninstalling the corresponding target artifacts.
```


achimweigel (approved):
/lgtm

PR #459, opened 2022-04-04
Update history

robertgraeff:

/area dev-productivity
/kind enhancement
/priority 3

Fixes #

**Release note**:

```other operator
NONE
```


robertgraeff:
Closing this pull request. The update history does not only describe the system, but writing the history triggers events that also influence the system. Previously, updates without any real change would have been without effect. With the added history every update would be a real change, resulting in endless reconciles.

PR #467, opened 2022-04-14
fix and improve helm tests

achimweigel:

/area helm-deployer
/kind enhancement
/priority 3

This PR updates the helm tests such that also value settings are tested. The local helm chart must be exchanged because the old one is not working anymore with newer kubernetes versions. 

Fixes #

**Release note**:

```other operator

```


reshnm (approved):
/lgtm

PR #470, opened 2022-04-21
handle the maximum length restriction for deployer registrations intstallation names more gracefully

reshnm:

/area deployers
/kind enhancement
/priority 3

The user can configure the "deployerManagment.agent.name" field in the landscaper helm chart values.
This value is used to create a Landscaper Environment. The landscaper then creates a installation for each registered deployer which is named by _$deployerRegistration-$environment_.
Landscaper installations have a maximum name length limitation of 58.
When the agent name is configured with a verly long name, this may prevent the landscaper to create the installation.

This PR basically does two things:

1. add an error to the landscaper helm chart when the value of "deployerManagment.agent.name" is longer than 30.
2. Add a Status to the DeployerRegistration CRD. The status contains a LastError field analog to other Landscaper CRDs. When the installation name for the DeployerRegistration exceeds the maximum length, The LastError is set accordingly.

**Release note**:

```other operator
The landscaper chart value "deployerManagment.agent.name" now has a maximum length of 30 to prevent failing installations of deployer registrations.
```


achimweigel (approved):
/lgtm

PR #483, opened 2022-05-05
Simplify reconcile helper

achimweigel:

/area deployers
/kind enhancement
/priority 3

Analysing the reconcile helper showed that most required data is already needed just after the object creation. Therefore it was possible to move this data creation from the requirements framework into the constructor. The only left data "dependencies" were refactored into a parameter making the data flow more explicit and allows the removal of the requirements framework. 

Fixes #

**Release note**:

```other operator
NONE
```


Diaphteiros (approved):
/ltgm

PR #484, opened 2022-05-06
Reconcile diagrams

robertgraeff:

/area documentation
/kind task
/priority 3

This pull request contains diagrams for the reconcile flow of the installation controller.

Fixes #

**Release note**:

```other operator
NONE
```


docs/technical/installation_controller.md
```diff
@@ -0,0 +1,38 @@
+# Installation Controller
+
+The reconciliation flow of the installation controller is described in detail [here](./reconciliation_flow.md).
+It executes one of the following main operations:
+
+- a deletion,
+- a [standard reconcile](#standard-reconcile),
+- a [force reconcile](#force-reconcile),
```
enrico-kaack-comp, line 8: the path "imports satisfied? -yes-> return error seems odd and doesnt fit the description in reconciliation_flow.md, does it?

docs/technical/installation_controller.md
```diff
@@ -0,0 +1,38 @@
+# Installation Controller
+
+The reconciliation flow of the installation controller is described in detail [here](./reconciliation_flow.md).
+It executes one of the following main operations:
+
```
enrico-kaack-comp, line 5: I dont find the documentation of all the phases in this folder. it would be useful to have it here (direct or referenced), since the phases are used in the following flows and explanations.

docs/technical/installation_controller.md
```diff
@@ -0,0 +1,38 @@
...
+- a deletion,
+- a [standard reconcile](#standard-reconcile),
+- a [force reconcile](#force-reconcile),
+- or a [re-check of completed installation](#re-check-completed-installation).
```
enrico-kaack-comp, line 9: first condition: phase == combined phase of sub-objects.
Does it mean all sub-objects must have the same phase or shall they all be successful? If latter, pls fix.

docs/technical/installation_controller.md
```diff
@@ -0,0 +1,38 @@
...
+
+The diagrams below illustrate these flows. First, the decision which main operation will be executed: 
+
+![inst-main-reconcile](images/inst-main-reconcile.png)
+
+
+## Standard Reconcile
+
+![inst-standard-reconcile](images/inst-standard-reconcile.png)
```
enrico-kaack-comp, line 18: first condition: sub-objects busy: does it mean that their phase is progressing?
condition update allowed returns an error, does this map to a function call? Otherwise it would be consistent to make it a yes/no condition

enrico-kaack-comp (changes requested):

docs/technical/installation_controller.md
robertgraeff: You are right, I mixed up "yes" and "no". I have updated the image.

docs/technical/installation_controller.md
robertgraeff: I added a paragraph about the phases and what the "combined phase" is.

docs/technical/installation_controller.md
robertgraeff: I added a paragraph about the phases and what the "combined phase" is.

docs/technical/installation_controller.md
robertgraeff: Sub-objects busy: it can be a progressing phase, but also if the spec was updated and the controller has not yet reconciled (observed generation != generation). I.e. working, or there is work to do which was not yet started. 

Update allowed: actually at first I used yes/no. Then Achim pointed out that the [method UpdateAllowed](https://github.com/gardener/landscaper/blob/e08eafcb1f1320d8935f01d4d09dda7853263513/pkg/landscaper/installations/reconcilehelper/reconcilehelper.go#L89) does not return a boolean, but an error instead of no.

enrico-kaack-comp (approved):
Great clarification, thank you. LGTM

PR #486, opened 2022-05-13
Full reconcile

robertgraeff:

/area robustness
/kind bug
/priority 3

If an installation has phase `Failed`, a manual action is necessary to trigger a new reconciliation that might solve the problem.

A force-reconcile annotation might be too brutal, because it skips some checks, so that the order of sub-installations / deploy items might not be guaranteed. 

The normal reconcile annotation on the other hand, is currently too weak. The reconciliation return quite early, because `Failed` is considered as a "completed" phase. The present pull request changes how the controllers handle objects with a reconcile annotation.  

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #116, opened 2021-02-11
Targetselector should look for values, not Values

MrBatschner:
**How to categorize this PR?**

/area quality
/kind cleanup
/priority normal

**What this PR does / why we need it**:

The Targetselector currently matches annotations through `key`, `operator` and `Values` (not the capital _V_ in `Values`). This should be a lowercase `values` (with a lowercase _v_).

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

PR #508, opened 2022-07-06
Landscaper deployment on kubernetes 1.24

robertgraeff:

/area quality
/kind bug
/priority 3

This pull request fixes the issue "Landscaper deployment does not install deployers #116"

In kubernetes version 1.24, no secret with a service account token is automatically generated for a service account. This has an impact on the generation of target objects that are used for the installation of deployers, because each of the targets contains a kubeconfig with a token originating from a service account. Affected service accounts: `container-default`, `helm-default`, `manifest-default`, `deployer-default`. The landscaper agent has corresponding service accounts per environment.  

This PR creates the secrets for the service accounts and waits until they are filled with a token. (The old behaviour was to wait for the secrets.)

Fixes #

**Release note**:

```other operator

```


achimweigel (approved):
/lgtm

PR #509, opened 2022-07-06
Check for duplicate exports

### LINKED ISSUE #507: duplicate export names are not prevented

/area robustness
/kind bug
/priority 1

**What happened**:
It is possible to create multiple installations which export the same dataobject. I just copied an existing installation which does nothing except for importing and exporting values and renamed all exports except for one data export.
The second installation succeeded and both installations were listed as owners in the conflicting `DataObject`. This might have been a special case, since both exported the same value, but it still should be prevented.

The second installation should not run at all and it should definitely not become `Succeeded`.

I used the manifests in `test/integration/installations/testdata/test1` for my test.

The whole Landscaper concept relies on the fact that export names are unique and it is clear which export comes from which installation. If this requirement is broken, the dependency management, and subsequently also the landscaper, will not work at all.

**Environment**:

- Landscaper version: `v0.26.0`

Diaphteiros:

/area robustness
/kind technical-debt
/priority 1

This PR adds the much required checks that a newly created/updated installation does not declare any export, which is already declared by another installation. This is essential for the Landscaper concept.

Fixes #507 

**Release note**:

```bugfix user
Landscaper now verifies that a created/updated installation does not declare exports which are already exported by another installation in the same context. This is checked in the validation webhook as well as during the reconciliation (in case the webhook is disabled). Installations which break this constraint will be denied by the webhook and set to `Failed` during the reconciliation.
```


Diaphteiros:
The reason for the failed test is solved with #511

pkg/landscaper/controllers/installations/controller.go
```diff
@@ -480,3 +498,26 @@ func (c *Controller) setInstallationPhaseAndUpdate(ctx context.Context, inst *ls
 
 	return lsError
 }
+
+func (c *Controller) checkForDuplicateExports(ctx context.Context, inst *lsv1alpha1.Installation) error {
```
achimweigel, line 502: With the new reconcile logic this works well for subinstallations but with respect to root installations I see problems with race conditions, e.g. if root inst RI-A is just creating its export data including export data EX and during that its spec is modified such that the new spec does not export EX anymore. This does not stop the export data generation of RI-A. If now another root inst RI-B is created also exporting EX and it checks the exports of RI-A is only sees the new spec without EX but EX is currently created resulting in potential problems. Or do I miss here something? 

Perhaps another strategy is to fail working on an installation if it tries to create some export data and this already exists and was created by another installation. Then we do not need a global check and would have no race conditions.

achimweigel (changes requested):

pkg/landscaper/controllers/installations/controller.go
Diaphteiros: You are right, this could happen, although it is somewhat unlikely.
I will add an additional check to the export creation. While only that check would be enough, I don't like the idea of detecting this problem during export creation, so I will keep the other checks in.

pkg/landscaper/controllers/installations/controller.go
```diff
@@ -141,6 +142,14 @@ func (c *Controller) reconcileNew(ctx context.Context, req reconcile.Request) (r
 
 	}
 
+	if inst.Status.ObservedGeneration != inst.Generation {
+		// spec has changed, make sure the exports don't conflict
+		err := c.checkForDuplicateExports(ctx, inst)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
```
achimweigel, line 151: Is there a reason why you not have removed the coding here?

pkg/landscaper/controllers/installations/controller.go
achimweigel: ok

pkg/landscaper/controllers/installations/reconcile.go
```diff
@@ -158,6 +158,11 @@ func (c *Controller) handleReconcilePhase(ctx context.Context, inst *lsv1alpha1.
 func (c *Controller) handlePhaseInit(ctx context.Context, inst *lsv1alpha1.Installation) (lserrors.LsError, lserrors.LsError) {
 	currentOperation := "handlePhaseInit"
 
+	err := c.checkForDuplicateExports(ctx, inst)
+	if err != nil {
+		return nil, lserrors.BuildLsError(err, currentOperation, "CheckForDuplicateExports", err.Error(), lsv1alpha1.ErrorConfigurationProblem)
```
achimweigel, line 163:
```suggestion
		return lserrors.BuildLsError(err, currentOperation, "CheckForDuplicateExports", err.Error(), lsv1alpha1.ErrorConfigurationProblem), nil
```

achimweigel: I would return this as a fatal error instead because manual intervention is required in that case.

achimweigel (changes requested):

achimweigel (approved):
/lgtm

PR #515, opened 2022-07-14
container deployer service account token handling

reshnm:

/area deployers
/kind enhancement
/priority 3

With this change, the container deployer creates the service account token secret manually, which is required for Kubernetes v1.24.x.
This PR also adds a requeue time for the container deployer garbage collector.
Previously objects (Service Accounts, Roles, Rolebindings etc.) created by the container deployer were not requed for reconciliation when the deploy item is still existing.

Fixes #

**Release note**:

```other user
For Kubernetes versions 1.24 and later, create the service account token secret manually in the container deployer.
Add container deployer garbage collector automatic requeuing.
```

### DIFF
--- a/apis/.schemes/container-v1alpha1-Configuration.json
+++ b/apis/.schemes/container-v1alpha1-Configuration.json
@@ -126,13 +126,20 @@
       "type": "object",
       "required": [
         "disable",
-        "worker"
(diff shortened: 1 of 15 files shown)


achimweigel (approved):
/lgtm

PR #534, opened 2022-08-02
added timed wait for objects, created by the integration test, being deleted

reshnm:

/area testing
/kind enhancement
/priority 3

Currently for some deployments created by the integration test, we don't wait for the actual deleteion but only for the deletion timestamp being set.
This could lead to race conditions in the integration tests.
This PR adds a timed wait for these objects being actually being deleted.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/integration/core/registry.go
+++ b/test/integration/core/registry.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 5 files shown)


Diaphteiros (approved):
/lgtm

PR #536, opened 2022-08-03
Adjust deploy item refs in execution status

robertgraeff:

/area robustness
/kind bug
/priority 3

During the cleanup of orphaned deploy items it is more robust to proceed in the following order (currently it is the reversed order):
1. cleanup deploy item references in the execution status  
2. delete the deploy item  

Moreover, the `seed` an `failFast` Ginkgo options are set to achieve that the integration tests
- always run in the same order,  
- and stop after the first test that fails.  

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #539, opened 2022-08-04
automatically generate documentation index

Diaphteiros:

/area documentation
/kind enhancement
/priority 5

This PR adds a small script which can automatically generate the documentation index (README.md in docs folder).

**How to use:**
Use `make generate-docs` to run the script. It is also included in `make generate`.

**How it works:**
The script first searches every folder in the `docs` folder for whether it contains a `.docnames` file. This is a meta file in JSON format, which contains the name with which this folder should be displayed in the index. Folders which don't have this file will be ignored. For all that have it, it then goes over all `*.md` files in them and generates links for the index.

The name for the link will be determined as follows:
Usually, the doc files have a fitting name set as header in the first line, which will be used. Alternatively, the name can be set manually by specifying it under `overwrites.<filename>` in the folder's meta file (see api reference for an example). If the first line of the file is empty and no overwrite is specified, the file will be ignored. To manually ignore a file, define an overwrite pointing to an empty string.

**Release note**:

```doc developer
The documentation index (README.md in the docs folder) can now be generated automatically by calling `make generate-docs`. It will also be generated on `make generate`.
```


achimweigel:
Would it perhaps make sense, to add the description in this PR how the index creation works to the developers documentation?

Diaphteiros:
> Would it perhaps make sense, to add the description in this PR how the index creation works to the developers documentation?

Valid point, will add it.

reshnm (approved):
Only minor nit, otherwise LGTM.

PR #541, opened 2022-08-05
Fix nilpointer in execution controller

robertgraeff:

/area robustness
/kind bug
/priority 3

This PR fixes a nilpointer in the execution controller. 
The error can occur after a force-delete operation and can cause a restart of the landscaper pod.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/landscaper/execution/classification.go
+++ b/pkg/landscaper/execution/classification.go
@@ -62,7 +62,12 @@ func newDeployItemClassification(executionJobID string, items []*executionItem) 
 	for i := range items {
 		item := items[i]
 
-		if item.DeployItem != nil && item.DeployItem.Status.JobID == executionJobID {
+		if item.DeployItem == nil {
+			// The items that we are classifying here were all created in the previous phase and should exist.
+			// But a user could have deleted items with "kubectl delete" or "landscaper-cli installations force-delete".
+			// We treat missing items as failed.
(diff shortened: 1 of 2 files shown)


pkg/landscaper/execution/classification.go
```diff
@@ -62,7 +62,11 @@ func newDeployItemClassification(executionJobID string, items []*executionItem)
 	for i := range items {
 		item := items[i]
 
-		if item.DeployItem != nil && item.DeployItem.Status.JobID == executionJobID {
+		if item.DeployItem == nil {
+			// The items that we are classifying here were all created in the previous phase and should exist.
+			// But for example a force-delete could have deleted them. We treat them as failed.
```
achimweigel, line 67: As far as I remember, we do not support force-delete with the new reconcile logic anymore

achimweigel: A problem with this approach is that the failed item are not part of the error message stored in the execution object. If the deploy item does not exist, it will be probably confusing for the user not to see any failed deploy item.

achimweigel (changes requested):

pkg/landscaper/execution/classification.go
robertgraeff: I have adjusted the texts as discussed.

achimweigel (approved):
/lgtm

PR #542, opened 2022-08-08
Fix computation of imports hash

robertgraeff:

/area robustness
/kind bug
/priority 3

This pull request fixes the following bugs concerning the hash value computation of the imports of an installation.

- The method that computes the hash value of a target list was not implemented (method `ComputeConfigGeneration` in the implementation of interface `ImportedBase` for target lists.)
- The hashes for different kinds of imports were all collected in the same map (due to a copy-paste error). We use now separate maps for dataobjects, targets, target lists etc.
- An alternative way to compute the imports hash is now deleted. This alternative was never used and was only in an experimental state.

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #543, opened 2022-08-09
Integration test script with cluster creation

robertgraeff:

/area testing
/kind enhancement
/priority 3

This pull request provides scripts to execute the Landscaper integration tests.

- script `.ci/new-integration-test-with-cluster-creation` creates a Gardener shoot cluster, then runs the tests using the shoot as test cluster, and finally deletes the shoot.
- script `.ci/new-integration-test` allows to run the execution tests using an existing cluster.

Fixes #

**Release note**:

```other operator
NONE
```


.ci/resources/cluster-template.yaml
```diff
@@ -0,0 +1,59 @@
+kind: Shoot
```
achimweigel, line 1: Why do we need this template anymore?

achimweigel (approved):
/lgtm

only one question if we really need one of the two templates

.ci/resources/cluster-template.yaml
robertgraeff: Removed it.

PR #546, opened 2022-08-09
Logging of the execution controller

robertgraeff:

/area logging
/kind task
/priority 3

This PR implements the new logging for the execution controller.

Fixes #

The utils/logging.go here is preliminary until the other pull requests for the logging are merged. 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/landscaper/controllers/execution/controller.go
+++ b/pkg/landscaper/controllers/execution/controller.go
@@ -10,36 +10,32 @@ import (
 	"reflect"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/gardener/landscaper/pkg/utils"
-
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-
-	"github.com/gardener/landscaper/controller-utils/pkg/kubernetes"
-	"github.com/gardener/landscaper/controller-utils/pkg/logging"
-
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	lserrors "github.com/gardener/landscaper/apis/errors"
 
 	lsv1alpha1 "github.com/gardener/landscaper/apis/core/v1alpha1"
 	lsv1alpha1helper "github.com/gardener/landscaper/apis/core/v1alpha1/helper"
+	lserrors "github.com/gardener/landscaper/apis/errors"
+	"github.com/gardener/landscaper/controller-utils/pkg/kubernetes"
 	kutil "github.com/gardener/landscaper/controller-utils/pkg/kubernetes"
+	"github.com/gardener/landscaper/controller-utils/pkg/logging"
+	lc "github.com/gardener/landscaper/controller-utils/pkg/logging/constants"
 	"github.com/gardener/landscaper/pkg/landscaper/execution"
 	"github.com/gardener/landscaper/pkg/landscaper/operation"
+	"github.com/gardener/landscaper/pkg/utils"
 	"github.com/gardener/landscaper/pkg/utils/read_write_layer"
 )
 
 // NewController creates a new execution controller that reconcile Execution resources.
-func NewController(log logging.Logger, kubeClient client.Client, scheme *runtime.Scheme, eventRecorder record.EventRecorder) (reconcile.Reconciler, error) {
+func NewController(logger logging.Logger, kubeClient client.Client, scheme *runtime.Scheme, eventRecorder record.EventRecorder) (reconcile.Reconciler, error) {
 	return &controller{
-		log:           log,
+		log:           logger,
 		client:        kubeClient,
 		scheme:        scheme,
 		eventRecorder: eventRecorder,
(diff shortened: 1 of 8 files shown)


pkg/landscaper/controllers/execution/controller.go
```diff
@@ -75,13 +76,13 @@ func (c *controller) reconcileNew(ctx context.Context, req reconcile.Request) (r
 
 	if exec.DeletionTimestamp.IsZero() && !kubernetes.HasFinalizer(exec, lsv1alpha1.LandscaperFinalizer) {
 		controllerutil.AddFinalizer(exec, lsv1alpha1.LandscaperFinalizer)
-		if err := c.Writer().UpdateExecution(ctx, read_write_layer.W000086, exec); err != nil {
+		if err := c.Writer(ctx).UpdateExecution(ctx, read_write_layer.W000086, exec); err != nil {
```
achimweigel, line 79: I'm not sure if I like this construct because it differs from the general pattern and fails if someone reuses a writer. 

```
w := c.Writer(ctx)
w.UpdateExecution()
....
w.UpdateExecution()
```

pkg/landscaper/controllers/execution/controller.go
```diff
@@ -439,33 +458,35 @@ func HandleAnnotationsAndGeneration(ctx context.Context, log logging.Logger, c c
 		// - force-reconcile annotation
 		// - outdated generation
 		opAnn := lsv1alpha1helper.GetOperation(exec.ObjectMeta)
-		log.Debug("reconcile required, setting observed generation and phase", "operationAnnotation", opAnn, "observedGeneration", exec.Status.ObservedGeneration, "generation", exec.Generation)
+		logger.Debug("reconcile required, setting observed generation and phase", "operationAnnotation", opAnn, "observedGeneration", exec.Status.ObservedGeneration, "generation", exec.Generation)
```
achimweigel, line 461: should be introduce constants for the keys and add them to the constants file in the logging package?

pkg/landscaper/execution/delete.go
```diff
@@ -90,7 +94,7 @@ func (o *Operation) checkDeletable(item executionItem, items []*executionItem) b
 		// the reconcile loop.
 		// Therefore, it should not be necessary to check again against the api server.
 		if exec.DeployItem != nil {
-			o.Log().Debug("deploy item %s depends on %s and is still present", exec.DeployItem.Name, item.Info.Name)
+			logger.WithValues("predecessor", item.Info.Name, "successor", exec.DeployItem.Name).Debug("deploy item still has a successor", exec.DeployItem.Name, item.Info.Name)
```
achimweigel, line 97: Should we remove the call of WithValues and add the key/value pairs to the Debug call? 

"exec.DeployItem.Name, item.Info.Name" is a strange key/value pair. Perhaps it would be better to add this info to the message?

pkg/landscaper/execution/reconcile.go
```diff
@@ -32,12 +34,14 @@ type executionItem struct {
 
 // Reconcile contains the reconcile logic for a execution item that schedules multiple DeployItems.
 func (o *Operation) Reconcile(ctx context.Context) lserrors.LsError {
+	logger, ctx := utils.FromContextOrNew(ctx)
+	logger = logger.WithValues("operation", "reconcile", "resource", kutil.ObjectKeyFromObject(o.exec).String())
```
achimweigel, line 38: we should use the constants for the keys from the logging package.

pkg/landscaper/execution/reconcile.go
```diff
@@ -49,26 +53,28 @@ func (o *Operation) Reconcile(ctx context.Context) lserrors.LsError {
...
 				return nil
 			} else if notCompletedPhaseOrNotStatusUpToDate(item.DeployItem) {
 				// deployitem is running - either in a non-final phase, or its observedGeneration doesn't match its generation
-				setPhaseProgressingOfRunningDeployItem(o.exec, item.Info.Name, o.EventRecorder(), dlogger)
+				setPhaseProgressingOfRunningDeployItem(diCtx, o.exec, item.Info.Name, o.EventRecorder())
 				allSucceeded = false
 			} else {
 				// the deployitem is: up-to-date, in a final state, not failed => deployItem.spec.phase == succeeded => nothing to do with the deployitem
-				dlogger.Debug("deployitem not triggered because up-to-date", "deployItemPhase", string(item.DeployItem.Status.Phase))
+				diLogger.Debug("deployitem not triggered because up-to-date", "deployItemPhase", string(item.DeployItem.Status.Phase))
```
achimweigel, line 77: do we need a constant for deployItemPhase?

achimweigel (changes requested):

pkg/landscaper/execution/delete.go
Diaphteiros: `log.WithValues("foo", "bar").Info("msg")` is the same as `log.Info("msg", "foo", "bar")`, so this can be simplified.

Regarding the second point, I agree that we shouldn't use the name of the DeployItem as key, since it is not deterministic and will only mess up our opensearch filtering. For the same reason, we shouldn't put the name as part of the message - it's easier to filter for a certain log message if the message is always the same. I think we can omit the `exec.DeployItem.Name, item.Info.Name` key-value-pair, because it contains the same information which is already given via the key-value-pairs in the `WithValues` part.

pkg/landscaper/execution/reconcile.go
Diaphteiros: There is actually a helper function for logging the beginning of a reconcile. Depending on whether you already have a logger or not, it is called as `logging.StartReconcileFromContext` or `logger.StartReconcile`. It takes a `reconcile.Request` as argument, which isn't present here, but maybe we can fake it?

The mentioned helper function doesn't log the operation, you can use `WithValues` for that. The `constants` package should already contain a key for `operation`.

pkg/landscaper/execution/reconcile.go
Diaphteiros: There are constants for the `Status.Phase` field for each resource (Installation, Execution, DeployItem). I didn't add one for the new deployitemPhase, because I didn't know how I should name it. Also, I personally would prefer to get rid of that field again, because I find it really confusing to have two phases with nearly the same purpose and no obvious distinction in the DeployItem status.

pkg/landscaper/controllers/execution/controller.go
robertgraeff: I have reverted this, so that the `Writer` method has no argument.

pkg/landscaper/execution/reconcile.go
achimweigel: Hi @Diaphteiros, Robert and I already planned to remove the additional phase field but perhaps in another PR.

pkg/landscaper/execution/delete.go
robertgraeff: Done

pkg/landscaper/execution/reconcile.go
robertgraeff: I changed this and log the method name using the `KeyMethod` constant.

pkg/landscaper/execution/reconcile.go
robertgraeff: I am using now the `KeyDeployItemPhase` constant.

achimweigel (approved):
/lgtm

PR #548, opened 2022-08-10
refactor FromContextOrNew method and move it to logging package

Diaphteiros:

/area logging
/kind enhancement
/priority 3

I changed the `FromContextOrNew` function in the following way:

The function now takes two different sets of key-value pairs. The first one is a fallback, for when no logger can be retrieved from the context and a new one has to be created. The second set will always be added to the logger. The disadvantage of this is that the arguments to the function now look slightly more ugly. The advantage is that no `logger.WithValues` and `logging.NewContext` calls after this function are required anymore, if one wants to add new key-value-pairs to the logger (which should also be put into the context).

Since we usually use this function in methods where we know that the context contains a logger, failing to get it now logs an Error. This makes the `CreatedBy` key-value-pair superfluous.

I also moved the method to the logging package, because I think that's the better place for it.

Additionally, I added an `FromContextWithFallback` method, which can be used if there is an existing logger (e.g. from the struct the method is called on), but one wants to check the context for a more specific one. The function tries to fetch the logger from the context, but in case of failure, it silently returns the provided fallback logger instead.

**Release note**:

```other operator
NONE
```


Diaphteiros:
@achimweigel @robertgraeff I made some changes to your `FromContextOrNew` function. What do you think?

achimweigel (approved):
/lgtm

I really like our new logging :-)

PR #551, opened 2022-08-11
Update deployer info in DI status

robertgraeff:

/area deployers
/kind bug
/priority 3

When a deployer processes a deploy items, it writes its own name and version into the status of the processed deploy item (field `.status.deployer`.) However, the version field was not updated, even though the deployer was running in the meantime in a newer version. So it appeared as if deploy items had been processed by an older deployer than was actually the case.

The present pull request fixes this.
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/deployer/lib/controller.go
+++ b/pkg/deployer/lib/controller.go
@@ -10,17 +10,10 @@ import (
 	"reflect"
 	"time"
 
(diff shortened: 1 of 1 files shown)


achimweigel (approved):
/lgtm

PR #552, opened 2022-08-11
improve logging in cmd package and agent

Diaphteiros:

/area logging
/kind enhancement
/priority 3

Improves the logging in the `cmd` package as well as the agent.
Also fixes two minor bugs in the logging framework which cause the landscaper to panic shortly after the start.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/cmd/landscaper-agent/app/app.go
+++ b/cmd/landscaper-agent/app/app.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
@@ -27,12 +26,10 @@ func NewLandscaperAgentCommand(ctx context.Context) *cobra.Command {
 
 		Run: func(cmd *cobra.Command, args []string) {
 			if err := options.Complete(); err != nil {
-				fmt.Print(err)
-				os.Exit(1)
+				panic(err)
 			}
 			if err := options.run(ctx); err != nil {
-				options.log.Error(err, "unable to run landscaper controller")
-				os.Exit(1)
+				panic(fmt.Errorf("unable to run landscaper controller: %w", err))
 			}
 		},
 	}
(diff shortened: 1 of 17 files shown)


cmd/landscaper-agent/app/app.go
```diff
@@ -43,7 +40,7 @@ func NewLandscaperAgentCommand(ctx context.Context) *cobra.Command {
 }
 
 func (o *options) run(ctx context.Context) error {
-	o.log.Info(fmt.Sprintf("Start Landscaper Agent with version %q", version.Get().String()))
+	o.log.Info("Starting Landscaper Agent", "version", version.Get().String())
```
achimweigel, line 43: Is there a reason why you do not use lc.KeyVersion here?

pkg/agent/agent.go
```diff
@@ -229,17 +252,22 @@ func (a *Agent) RemoveHostResources(ctx context.Context, kubeClient client.Clien
 
 // Reconcile reconciles the environment and target on the landscaper.
 func (a *Agent) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	logger, ctx := logging.MustStartReconcileFromContext(ctx, req, nil)
 	if req.Name != a.config.Name {
+		logger.Info("Not responsible for this environment", "targetEnvironment", a.config.Name)
```
achimweigel, line 257: Do we need a constant for targetEnvironment?

achimweigel (approved):
/lgtm

on two small questions

cmd/landscaper-agent/app/app.go
Diaphteiros: No, maybe that was one of the first changes before the `KeyVersion` field was added. I will update it.

pkg/agent/agent.go
Diaphteiros: Probably not, I don't think this key will appear more than once. Hm, maybe we don't even need it here, I think I add the name of the environment the agent is responsible for already to the values at a different point. Will check.

PR #556, opened 2022-08-12
Improve logging for controller-utils

reshnm:

/area logging
/kind enhancement
/priority 3

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/controller-utils/pkg/kubernetes/kubernetes.go
+++ b/controller-utils/pkg/kubernetes/kubernetes.go
@@ -532,7 +532,7 @@ func DecodeObjects(log logging.Logger, name string, data []byte) ([]*unstructure
 			if err == io.EOF {
 				break
 			}
-			log.Error(err, fmt.Sprintf("unable to decode resource %d of file %q", i, name))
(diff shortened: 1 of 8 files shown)


controller-utils/pkg/kubernetes/kubernetes.go
```diff
@@ -532,7 +532,7 @@ func DecodeObjects(log logging.Logger, name string, data []byte) ([]*unstructure
 			if err == io.EOF {
 				break
 			}
-			log.Error(err, fmt.Sprintf("unable to decode resource %d of file %q", i, name))
+			log.Error(err, "unable to decode resource", "resourceIndex", i, "fileName", name)
```
achimweigel, line 535: I'm not sure if we need constants for index and fileName?

achimweigel (approved):
/lgtm

only one small question

controller-utils/pkg/kubernetes/kubernetes.go
reshnm: Oh, I think that makes sense. Especially file name.

PR #563, opened 2022-08-17
remove the check for an manifest/helm export to be a managed resource

reshnm:

/area deployers
/kind enhancement
/priority 3

Previously for exports in helm/manifest deployitems, the resource from which the value is exported, was checked for being managed by the deployer. This only worsens the usability without serving any valid use case.
This PR removes the check for being a manged resource.

Fixes #

**Release note**:

```feature user
Helm and Manifest Deployer can now epxport data from non-managed resources.
```

### DIFF
--- a/docs/deployer/helm.md
+++ b/docs/deployer/helm.md
@@ -130,45 +130,39 @@ spec:
     createNamespace: true
     # Values to template the chart
     # optional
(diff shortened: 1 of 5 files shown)


achimweigel (approved):
/lgtm

PR #574, opened 2022-08-29
Fix update of target import

robertgraeff:

/area quality
/kind bug
/priority 3

This pull requests fixes the following issue concerning the update of a target import.

Apply an installation which has a subinstallation and imports a Target which is passed to the subinstallation.
Wait until the reconciliation of the installation has finished (i.e. until the phase is Succeeded or Failed).
During the reconcile, a copy of the target is created in the "context" of the installation.

Update the target (for example the token of the kubeconfig). Trigger a new reconciliation with the reconcile annotation.

The new reconciliation does not update the copy of the target, but it should do this. As a consequence, the deploy items of the subinstallation will still use the old kubeconfig, which is wrong.

What you expected to happen:

How to reproduce it (as minimally and precisely as possible):

[This example](https://github.com/gardener/landscaper-examples/tree/master/subinstallations/export-import/v0.1.2) can be used to reproduce the issue.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/landscaper/dataobjects/target.go
+++ b/pkg/landscaper/dataobjects/target.go
@@ -85,36 +85,27 @@ func (t *Target) SetOwner(own *metav1.OwnerReference) *Target {
 	return t
 }
 
-// Build creates a new data object based on the given data and metadata.
-// Does not set owner references.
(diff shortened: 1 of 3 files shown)


achimweigel (approved):
/lgtm

PR #582, opened 2022-09-02
Refactor target extension

achimweigel:

/area deployers
/kind enhancement
/priority 3

Refactor target extension plus

- remove old reconcile logic for installations

Fixes #

**Release note**:

```other operator
NONE
```


reshnm:
@achimweigel I'm not sure whether I'm happy with the naming in general.
For installations we have an internal `Installation` struct containing an `InstallationBase` struct.
This seems very similar to the `TargetExtension` that is used here. And I don't know whether this is the only place.

I would like to have similar names for similar things. So either also use the extension naming pattern also for installations or use a `TargetBase`. What do you think?

achimweigel:
@reshnm I agree with you that this strange installation containing an installation base containing an "real" installation needs to be refactored too but this will be an even more complicated step. 

My problem with `Base` is, that this indicates something less but I wanted to express here that a `TargetExtension` contains a target and more. Robert and I were not totally happy with `Extension` but we searched around for a while and did not find anything better. I have no problem if we find something better but it must express that the type includes more and not less. The best would be to find names reflecting the semantics of the types but so far I was not able to figure this out.

reshnm:
@achimweigel To be clear, I'm totally fine with `Extension`. I just would like keep this naming for other similar things, like the Installation example. So maybe in the future we also change the internal Installation to `InstallationExtension`.

achimweigel:
@reshnm got it. Yes that it exactly what we should do. We should use a unique naming for something which is more than something else and for which we could not found a good name describing the semantics.

robertgraeff (approved):

PR #602, opened 2022-09-22
fix local integration tests

Diaphteiros:

/area testing
/kind enhancement
/priority 3

This PR fixes some issues which prevented me from running the integration tests locally:
- The script which installs the landscaper didn't take the current version (including git hash) into account, causing it to deploy the landscaper with image version `v0.36.0-dev`, which doesn't exist, instead of `v0.36.0-dev-<hash>`, which is the version created by the Makefile.
- The script which creates the registry for integration tests can be configured to use a loadbalancer service for the registry, so that it is reachable from outside of the cluster. However, it only tried to fetch an IP from the service's status, but on AWS, loadbalancers don't have IPs but hostnames instead. This caused the script to fallback to an internal DNS name and made the registry integration tests unable to run locally when targeting a cluster running on AWS.
- At least for the AWS case with the loadbalancer hostname, the registry creation script has to wait until the DNS lookup for the newly generated loadbalancer works, otherwise the integration tests will fail because the registry cannot be reached.
- It's now possible to run the integration tests without installing Landscaper or registry and without checking whether the Landscaper pods are running. This is useful if one wants to debug e.g. a deployer and therefore run this deployer locally instead of in the cluster. This mode does not support tests which require a registry at the moment.

**Release note**:

```other developer
Improved the registry creation script which is used for the integration tests to be more robust.
```
```breaking developer
The `./hack/testcluster registry create` command, which is used for the integration tests, now takes a `--dns-format` argument with the possible values `internal` and `external`. This replaces the old `--address-format` argument. The old value `ip` corresponds to the new `external`, while the old `hostname` corresponds to the new `internal`. Note that, while the actual functionality didn't change too much, the new `external` mode will not always result in an IP, but might also result in a hostname (hence the name change).
```
```feature developer
A new make target `integration-test-pure` has been added. If run, it executes the integration tests, but it does neither deploy the Landscaper into the cluster (has to be there already), nor a registry (tests which require it are not supported currently). Also, the check whether the Landscaper pods are ready is skipped. This is intended for cases where a certain Landscaper component (e.g. a deployer) is run locally to debug into it and should therefore not be active in the cluster.
```

### DIFF
--- a/.ci/int-test-helper/create-registry
+++ b/.ci/int-test-helper/create-registry
@@ -47,4 +47,22 @@ go run -mod=vendor ./hack/testcluster registry create \
     --timeout=10m \
     --ls-run-on-shoot
 
+address=$(cat "$TMP/docker.config" | jq -r '.auths | to_entries | .[0].value.serveraddress')
+address=${address%:*} # remove port, if given
+echo "Waiting until registry address '$address' is routable ..."
+timeout=300
+start=$(date +%s)
+now=start
+while true; do
+  if nslookup "$address" >/dev/null; then
+    # successfully routed
+    break
+  elif [[ $(($now - $start)) -gt $timeout ]]; then
+    echo "Address '$address' did not become routable within $timeout seconds."
+    exit 1
+  fi
+  sleep 5
+  now=$(date +%s)
+done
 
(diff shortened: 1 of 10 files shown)


achimweigel (approved):
/lgtm

cool stuff with AWS, I was not aware of  :-)

Issue #600, opened 2022-09-21
CustomReadinessCheck for Helm Deployer not working


enrico-kaack-comp:
**How to categorize this issue?**

/area helm-deployer
/kind bug
/priority 3

**What happened**:
Using a custom readiness check with the real helm deployers does not work. It always succeed.

**How to reproduce it (as minimally and precisely as possible)**:
use a simple helm chart and add a readiness check. Set ProviderConfiguration.HelmDeployment to true.

**Anything else we need to know?**:
The Custom ReadinessCheck seems to report success in this early exit:
https://github.com/gardener/landscaper/blob/master/pkg/deployer/lib/readinesscheck/customreadinesscheck.go#L44

`len(c.ManagedResources) == 0` is true, it seems the ManagedResources are not correctly tracked by the real helm deployer.

**Environment**:

- Landscaper version: v0.35.0-dev
- Kubernetes version (use `kubectl version`): v1.24.3
- Others:

achimweigel:
@enrico-kaack-comp could this issue be closed?

PR #616, opened 2022-10-11
Fix readiness check helm deployer

enrico-kaack-comp:

/area helm-deployer
/kind bug
/priority 3

Fixes the malfunctioning readiness check when using the real helm deployer #600 

Fixes #

**Release note**:

```bugfix user
- fix readiness check not working when using the real helm deployer 
```

### DIFF
--- a/pkg/deployer/lib/readinesscheck/readiness.go
+++ b/pkg/deployer/lib/readinesscheck/readiness.go
@@ -10,7 +10,6 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apimacherrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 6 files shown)


pkg/deployer/lib/readinesscheck/readiness.go
```diff
@@ -108,18 +113,25 @@ func (e *ObjectNotReadyError) Error() string {
...
+	}
+
 	key := kutil.ObjectKey(obj.GetName(), obj.GetNamespace())
 	if err := kubeClient.Get(ctx, key, obj); err != nil {
-		objLog.Debug("Resource status", lc.KeyStatus, StatusUnknown)
-		return fmt.Errorf("unable to get %s %s/%s: %w",
-			obj.GroupVersionKind().String(),
-			obj.GetName(), obj.GetNamespace(),
-			err)
+		if !errors.IsNotFound(err) || isCheckRelevant(obj) || failOnMissingObject {
```
achimweigel, line 128: as we have the check in line 122 `if !isCheckRelevant(obj) && !failOnMissingObject {` we could remove the complete check here in line 128 `if !errors.IsNotFound(err) || isCheckRelevant(obj) || failOnMissingObject {`

achimweigel (changes requested):

pkg/deployer/lib/readinesscheck/readiness.go
robertgraeff: Correct.

This means we don't need to pass the `isCheckRelevant` function and the `failOnMissingObject` flag through all the methods. We could simply filter the object list earlier, and call `WaitForObjectReady` for the already filtered object list.

achimweigel (approved):
/lgtm

PR #623, opened 2022-10-20
add message to '.Error()' output for LsError

Diaphteiros:

/area logging
/kind enhancement
/priority 3

When wrapping an `error` in an object of type `LsError`, a message can be specified. Calling `.Error()` on the LsError object returns only the original error message, though. If this LsError object is then again wrapped, neither `.Message`, nor `.Error()` called on the new object will return the message specified during the first wrapping. This results in a loss of potentially interesting information.

To fix this, the `.Error()` implementation of LsError has been enhanced in the `LsError` type: Instead of just returning `.Error()` of the internal `error`, it will now return a combination of `.Message` and `.Error()`. It also checks if former one already contains the latter one, to avoid duplicate errors.

**Release note**:

```other developer
Calling `.Error()` on a `LsError` object will now return a combination of the LsError's message and its internal error, instead of just returning the internal error's message.
```

### DIFF
--- a/apis/errors/error.go
+++ b/apis/errors/error.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -31,7 +32,15 @@ type Error struct {
 // Error implements the error interface
 func (e Error) Error() string {
 	if e.err != nil {
-		return e.err.Error()
+		internalErr := e.err.Error()
+		if len(e.lsErr.Message) > 0 {
+			if !strings.HasSuffix(e.lsErr.Message, internalErr) {
+				// if the developer has not already included the original error in the message, let's add it
+				return fmt.Sprintf("%s: %s", e.lsErr.Message, internalErr)
+			}
+			return e.lsErr.Message
+		}
+		return internalErr
 	}
 	return fmt.Sprintf("Op: %s - Reason: %s - Message: %s", e.lsErr.Operation, e.lsErr.Reason, e.lsErr.Message)
 }
--- a/apis/errors/error.go
+++ b/apis/errors/error.go
@@ -31,18 +31,19 @@ type Error struct {
 
 // Error implements the error interface
 func (e Error) Error() string {
+	message := e.lsErr.Message
 	if e.err != nil {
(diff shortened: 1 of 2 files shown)


Diaphteiros:
cc @reshnm

apis/errors/error.go
```diff
@@ -31,7 +32,15 @@ type Error struct {
 // Error implements the error interface
 func (e Error) Error() string {
 	if e.err != nil {
-		return e.err.Error()
+		internalErr := e.err.Error()
+		if len(e.lsErr.Message) > 0 {
+			if !strings.HasSuffix(e.lsErr.Message, internalErr) {
+				// if the developer has not already included the original error in the message, let's add it
+				return fmt.Sprintf("%s: %s", e.lsErr.Message, internalErr)
```
achimweigel, line 39:
```suggestion
				return return fmt.Sprintf("Op: %s - Reason: %s - Message: %s", e.lsErr.Operation, e.lsErr.Reason, fmt.Sprintf("%s: %s", e.lsErr.Message, internalErr))
```

I'm not sure if I really completely understood the discussion but aren't we loosing the operation and the reason if there is an internal error?

Diaphteiros: Yes. But I didn't change that, it was already implemented the way that operation and reason are only returned if there is no internal error.
In the installation resource, the reason and operation fields are read, if the error is a LsError, and put into the respective fields of the installation's status. So this is only a problem for the logs, I guess ...

apis/errors/error.go
```diff
@@ -31,7 +32,15 @@ type Error struct {
...
+			}
+			return e.lsErr.Message
+		}
+		return internalErr
```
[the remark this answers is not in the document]
Diaphteiros: Same reason as above - I kept the original implementation for this case.

apis/errors/error.go
Diaphteiros: But I agree with you that it would be nice to have the information for the logs too. Only ugly thing is that it will be duplicated in the installation's status, if we dump it into the error message, but that shouldn't be much of a problem.
What do you think?

achimweigel: I would prefer to start with providing more information even if there are duplicates. We could still optimise long or unneeded information later but I think missing info in a support case is more critical.

achimweigel (approved):
/lgtm

PR #624, opened 2022-10-25
add new manifest deployer update strategies

reshnm:

/area manifest-deployer
/kind enhancement
/priority 3

The previously available manifest deployer update strategies `update` and `patch` were not sufficient for resources that are being modified by a kubernetes controller after being created (like Gardener Shoots). The available strategies would always overwrite these modfications.

This PR introduces two new update strategies to accomodate the use case scenarios:

* `merge`: The manifest deployer will merge the results of the rendered manifests into the resources on the cluster. Fields that already exist in the resources on the cluster, will not be overwritten.
* `mergeOverwrite`: The manifest deployer will merge the results of the rendered manifests into the resources on the cluster. Fields that already exist in the resources on the cluster, will be overwritten when the rendered field is not empty.

Fixes #

**Release note**:

```feature user
Add manifest deployer update strategies "merge" and "mergeOverwrite".
```

### DIFF
--- a/apis/deployer/manifest/types_provider.go
+++ b/apis/deployer/manifest/types_provider.go
@@ -47,8 +47,10 @@ type ProviderConfiguration struct {
 type UpdateStrategy string
 
 const (
-	UpdateStrategyUpdate UpdateStrategy = "update"
-	UpdateStrategyPatch  UpdateStrategy = "patch"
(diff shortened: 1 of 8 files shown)


Diaphteiros (approved):
/lgtm

PR #627, opened 2022-10-26
add 'annotateBeforeDelete' and 'annotateBeforeCreate' to the manifest deployer

reshnm:

/area manifest-deployer
/kind enhancement
/priority 3

Some Kubernetes resources like Gardener Shoots need to be annotated before they can be deleted.
This PR adds an optional `annotateBeforeDelete` field to the manifest specification. All key, value pairs under this field are added as annotation before the manifest is being deleted.

For symmetry, the field `annotateBeforeCreate` is also added optionally to the manifest specification. These annotations are getting added to the manifest before it is being created on the target. These annotations are not getting re-applied on an update of that manifest. 

Fixes #

**Release note**:

```feature user
Add 'annotateBeforeDelete' and 'annotateBeforeCreate' manifest fields to the manifest deployer.
```


Diaphteiros (approved):
/lgtm

PR #628, opened 2022-10-26
fix manifest deployer customreadiness timeout

reshnm:

/area manifest-deployer
/kind bug
/priority 3

The custom readiness check timeout was not respected when set.
This is fixed by this PR.

Fixes #

**Release note**:

```bugfix user
Respect the custom readiness check timeout when set.
```


achimweigel (approved):
/lgtm

PR #629, opened 2022-10-27
resolve secret references in targets in deployer lib

Diaphteiros:

/area usability
/kind enhancement
/priority 3

- Target type definitions have moved to `apis/core/v1alpha1/targettypes`
- A new type `ResolvedTarget` has been added. It contains a Target and an additional `Content` field for the target's content. This type is used in the deployer library now instead of Target.
- The deployer library now resolves secret references in Targets before calling the deployer-specific functions.
- Target.Spec.Configuration is now `*AnyJSON` instead of `AnyJSON`.
- The webhook now validates that Targets contain either `Configuration` or `SecretRef` and not both.
- Target.Spec.SecretRef is now a `LocalSecretReference`, which takes only a `Name` and optionally a `Key`. `Namespace` is not supported anymore.

**Release note**:

```breaking developer
The definitions for Target types have moved from `apis/core/v1alpha1` to `apis/core/v1alpha1/targettypes`.
```
```breaking developer
The deployer library now resolves secret references in Targets (only those referenced in `Spec.Target` of the deployitem) before calling the deployer-specific functions. Instead of providing a `Target` to these functions, it now gives a `ResolvedTarget` as argument. This is a new type, consisting of the original Target as well as a `Content` field, which contains the actual content of the Target - either its inline configuration or the result of the resolved reference.
```
```breaking developer
Target's `Spec.Configuration` field now is of type `*AnyJSON` (instead of `AnyJSON`). It is expected to be `nil`, when `Spec.SecretRef` is set.
```
```breaking user
Target's `Spec.SecretRef` field now is of type `*LocalSecretReference` (instead of `SecretReference`). This means that a secret reference in a target may not contain a `Namespace` anymore, only `Name` and optionally `Key` are allowed.
```
```breaking user
Referencing a kubeconfig secret in a Target of type `landscaper.gardener.cloud/kubernetes-cluster` by specifying `kubeconfig.secretRef` is now deprecated and will be removed in the future. Use the Target's `spec.secretRef` field instead.
```
```feature operator
The Landscaper webhook now validates that at most one of `Configuration` and `SecretRef` in a Target's `Spec` is set, not both. It is possible to create empty Targets with neither field being set.
```
```feature user
The resolved Target referenced at `.spec.target` in a container deployitem is now available within the container. The `TARGET_PATH` env variable points to a json file containing the original Target (under `target`) as well as its resolved content (under `content`). See the [updated documentation](docs/deployer/container.md).
```
```other developer
The timeout durations for pickup and abortion timeouts in the integration tests have been increased from 10 seconds to 30 seconds to stabilize some tests.
```


apis/core/v1alpha1/targettypes/kubernetes_cluster.go
```diff
@@ -0,0 +1,94 @@
...
+	err := json.Unmarshal(data, kj)
+	if err == nil && kj.Kubeconfig != nil {
+		// parsing was successful
+		kc.Kubeconfig = *kj.Kubeconfig
+		return nil
+	}
+	return kc.Kubeconfig.UnmarshalJSON(data)
+}
+
+func (v ValueRef) OpenAPISchemaType() []string {
```
achimweigel, line 88: For what is this required? I just want to learn something :-)

apis/core/v1alpha1/targettypes/kubernetes_cluster.go
```diff
@@ -0,0 +1,94 @@
...
+	return []string{
+		"object",
+		"string",
+	}
+}
+func (v ValueRef) OpenAPISchemaFormat() string { return "" }
```
achimweigel, line 94: For what is this required?

docs/technical/target_types.md
```diff
@@ -34,24 +34,7 @@ spec:
...
-spec:
-    config:
-      secretRef:
-        name: my-secret
-        namespace: default
-        key: kubeconfig # optional will default to "kubeconfig"
-```
-
-**Config Variant 3**:
+**DEPRECATED: use the Target's `spec.secretRef` field instead**
```
achimweigel, line 37: Should we perhaps add a sentence like? The variant with a reference in the kubeconfig section if deprecated...

achimweigel (approved):
/lgtm

apis/core/types_target.go
```diff
@@ -41,12 +41,12 @@ type TargetSpec struct {
...
 	// Exactly one of the fields Configuration and SecretRef must be set
 	// +optional
-	Configuration AnyJSON `json:"config,omitempty"`
+	Configuration *AnyJSON `json:"config,omitempty"`
 
 	// Reference to a secret containing the target type specific configuration.
 	// Exactly one of the fields Configuration and SecretRef must be set
 	// +optional
-	SecretRef *SecretReference `json:"secretRef,omitempty"`
+	SecretRef *LocalSecretReference `json:"secretRef,omitempty"`
```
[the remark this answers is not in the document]
Diaphteiros: It's a breaking change, k8s will now deny creation of Targets which contain a namespace, I think. But you have added the support for top-level secret references only a few commits ago, so there shouldn't be too many Targets out there which use this feature.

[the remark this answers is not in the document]
achimweigel: that's true. I totally forgot :-)

apis/core/v1alpha1/targettypes/kubernetes_cluster.go
Diaphteiros: I'm not completely sure either, I just copied it. But I think it's related to the spec generation here https://github.com/gardener/landscaper/blob/master/docs/api-reference/core.md

apis/core/v1alpha1/targettypes/kubernetes_cluster.go
Diaphteiros: see above

docs/technical/target_types.md
Diaphteiros: Where do you want to add this sentence?

achimweigel: Perhaps below the deprecation hint?

Diaphteiros: I'd say we don't need another sentence, as the current hint seems pretty obvious to me, but I don't have a problem with adding one either. I'll add one.

achimweigel: I only thought, if a user sees a target with the old format, how should he know that this is variant 3, if this format is not mentioned anymore?

achimweigel: But if you do not agree, I'm totally fine with the current version.

PR #630, opened 2022-10-27
fix the continuous custom readiness check

reshnm:

/area deployers
/kind bug
/priority 3

The custom readiness check was not continously check for the readiness requirements to be fulfilled until the timeout was exceeded.
Instead the readiness check was aborted once the first requirement was not fulfilled.

Fixes #

**Release note**:

```bugfix user
Fix the custom readiness check for the Helm and Manifest deployers to wait for all contitions to be fulfilled until the timeout is exceeded.
```


achimweigel (approved):
/lgtm

PR #633, opened 2022-11-08
Test reconcile annotation

robertgraeff:

/area testing
/kind test
/priority 3

This pull request introduces the following annotation for DeployItems:

```yaml
metadata:
  annotations:
    landscaper.gardener.cloud/operation: test-reconcile
```

 If the annotation is set, the DeployItem will get a new `jobID` so that it will be reconciled.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/deployer/lib/controller.go
+++ b/pkg/deployer/lib/controller.go
@@ -188,7 +188,7 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			return reconcile.Result{}, err
 		}
 
-		logger.Error(nil, "generating a new jobID, because of a test-reconcile annotation")
+		logger.Info("generating a new jobID, because of a test-reconcile annotation")
 		di.Status.JobID = uuid.New().String()
 		if err := c.Writer().UpdateDeployItemStatus(ctx, read_write_layer.W000148, di); err != nil {
 			return reconcile.Result{}, err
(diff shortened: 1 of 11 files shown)


achimweigel (approved):
/lgtm

pkg/deployer/lib/controller.go
```diff
@@ -180,6 +182,21 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, fmt.Errorf("unable to get landscaper context: %w", err)
 	}
 
+	if lsv1alpha1helper.HasOperation(di.ObjectMeta, lsv1alpha1.TestReconcileOperation) {
+
+		if err := c.removeTestReconcileAnnotation(ctx, di); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		logger.Error(nil, "generating a new jobID, because of a test-reconcile annotation")
```
Diaphteiros, line 191: Why do we log this on error level? Shouldn't it be info?

pkg/deployer/lib/controller.go
```diff
@@ -292,6 +310,18 @@ func (c *controller) updateDiForNewReconcile(ctx context.Context, di *lsv1alpha1
 	return nil
 }
 
+func (c *controller) removeTestReconcileAnnotation(ctx context.Context, di *lsv1alpha1.DeployItem) lserrors.LsError {
+	logger, ctx := logging.FromContextOrNew(ctx, []interface{}{lc.KeyReconciledResource, client.ObjectKeyFromObject(di).String()})
+
+	logger.Error(nil, "remove test-reconcile annotation: this is only an error if this happens on a productive system")
```
Diaphteiros, line 316: see above

Diaphteiros (commented):
I was a little bit confused about the log levels. Looks good otherwise.

I think I get your train of thoughts here: it is logged as error, because it is something which shouldn't happen on a productive system.
I would argue that this is never unexpected behavior, because it always has to be triggered by a user. Therefore it should be logged on info level.

But this is just a minor nit. It's not like we are going to see this annotation often anyway, so I'm fine with it either way.

If we want to be really fancy, we could make this annotation work only for cases in which the deployitem does not have an owner - then it simply wouldn't work in productive use-cases, where the deployitems are created (and owned) by executions.

pkg/deployer/lib/controller.go
robertgraeff: I change the log level to `Info`.

pkg/deployer/lib/controller.go
robertgraeff: I change the log level to `Info`.

PR #635, opened 2022-11-09
Handle Failed Deletion of Subinstallations

robertgraeff:

/area quality
/kind bug
/priority 3

Suppose a root installation has two subinstallations A and B. Suppose B is a successor of A, i.e. B imports an export of A. Delete the root installation. Suppose the deletion of B fails (phase `DeleteFailed`). Then the deletion is stuck with A in phase  `InitDelete` and the root installation in phase `Deleting`. Neither an interrupt annotation, nor a reconcile annotation help.

The current pull request fixes this. If the deletion of B has failed, also A and the root installation will fail. Then the deletion can be repeated by adding a reconcile annotation at the root installation.

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #636, opened 2022-11-11
improve error message for duplicate sibling exports

Diaphteiros:

/area logging
/kind enhancement
/priority 3

**Release note**:

```other operator
The error message for duplicate exports among subinstallations now provides information on which exports from which subinstallations are conflicting.
```
```bugfix operator
Fixed a bug which caused the cyclic dependency check for subinstallations to sometimes contain the wrong cycle in the error message.
```
```other developer
The tests for cyclic dependency detection now validate that the correct cycles are found.
```


achimweigel (approved):
/lgtm

PR #646, opened 2022-11-25
enhance and document TargetList imports

Diaphteiros:

/area usability
/kind enhancement
/priority 3

**Release note**:

```feature user
It is now possible to use a single Target from a TargetList import of the parent Installation in a subinstallation. This allows, for example, to create one subinstallation per Target of a TargetList. See the [new TargetList documentation](docs/usage/TargetLists.md) for details and examples.
```
```doc user
A dedicated documentation for TargetList imports has been added, see [here](docs/usage/TargetLists.md).
```


achimweigel (approved):
/lgtm

The documentation is much better than before and really good now. I like the example snippets.

Only a question, if you have some simple running examples, what about adding them to our example repo and add a link to them to the docu. If not, that's also ok.

PR #656, opened 2022-12-09
Workaround for setup-envtest

robertgraeff:
**How to categorize this PR?**

/area robustness
/kind bug
/priority 3

**What this PR does / why we need it**:

Currently, `go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest` fails. We use this in `make install-requirements`, and therefore in the `verify` task of some pipeline jobs.

See issue https://github.com/kubernetes-sigs/controller-runtime/issues/2086

As a temporary workaround we can use a fixed version instead of `latest`.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
NONE
```

PR #657, opened 2022-12-12
Revert "Workaround for setup-envtest"

robertgraeff:
Reverts gardener/landscaper#656

The workaround is no longer be necessary, because the original issue is resolved.


achimweigel (approved):
/lgtm

PR #658, opened 2022-12-12
Enable spiff templating for inline text

Diaphteiros:

/area usability
/kind enhancement
/priority 5

Before this change, `GoTemplate` required an inline template to be a single string, while `Spiff` only worked with YAML/JSON. Now, an inline string is also accepted for Spiff.

**Before**
```yaml
- name: one
  type: Spiff
  template:
    deployItems:
    - name: init
      type: container
      config:
        apiVersion: example.test/v1
        kind: Configuration
        image: (( "my-custom-image:" imports.version ))
```
**Now also possible**
```yaml
- name: one
  type: Spiff
  template: |
    deployItems:
    - name: init
      type: container
      config:
        apiVersion: example.test/v1
        kind: Configuration
        image: (( "my-custom-image:" imports.version ))
```

**Release note**:

```enhancement user
When using `Spiff` templating, inline templates are now also accepted as text (similar to `GoTemplate`). Before this change, only YAML/JSON was possible.
```


achimweigel (approved):
/lgtm

PR #659, opened 2022-12-12
check that docs index is up-to-date during verify

Diaphteiros:

/area dev-productivity
/kind enhancement
/priority 3

During `make verify`, it is now also verified that the documentation index is up-to-date.

**Release note**:

```other developer
`make verify` now also verifies that the documentation index is up-to-date.
```
```other developer
`make install-requirements` now also installs `jq` (if it isn't already installed), which is needed for generating the documentation index. When run with sufficient privileges, it will copy the binary into `usr/local/bin`, otherwise it will add a temporary folder to the `PATH` env var.
```


achimweigel (approved):
/lgtm

PR #665, opened 2023-01-10
Check Interruption of Readiness Check and Export Collection

robertgraeff:

/area robustness
/kind bug
/priority 3

Currently, the helm and manifest deployers wait during a readiness check and during the collection of export values. (This behaviour is not changed by this pull request.) This pull request adds a check whether the deploy item was interrupted in the meantime. In this case the readiness check (resp. export collection) is stopped.

Without this fix, an interrupted deploy item might appear finished (phase = Failed), although the deployer is still processing it. If you then trigger a new reconcile, the new reconcile of the deploy item does not start until the current reconcile has really finished, for example due to a timeout.

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #677, opened 2023-01-26
use minimalistic helm chart instead of nginx chart for helm deployer test

Diaphteiros:

/area testing
/kind enhancement
/priority 3

Instead of deploying the nginx helm chart from the tutorial, the helm deployer integration test will now use the minimalistic helm chart from [here](https://github.com/gardener/landscaper-examples/tree/master/hello-world).
Some failures of the integration test seem to not come from any problems within the Landscaper, but from helm itself. Reducing the complexity of the used helm chart will hopefully increase the stability and reduce false positives of the integration tests.

Disadvantage: This creates kind of a dependency between the Landscaper and the Landscaper examples repo. On the other hand: The uploaded helm chart, which is now referenced in the integration test, won't change in the OCI registry, so incompatibilities are only expected if there are breaking changes in the Landscaper itself, and then we want to update the examples anyway, so I don't see too much of a problem here.

**Release note**:

```other developer
The deployer blueprint integration tests now use a minimalistic helm chart for verifying the helm deployer instead of the nginx helm chart from the tutorial. This is supposed to improve stability of the integration tests.
```

### DIFF
--- a/test/integration/deployers/blueprints/blueprints_tests.go
+++ b/test/integration/deployers/blueprints/blueprints_tests.go
@@ -94,10 +94,7 @@ func DeployerBlueprintTests(f *framework.Framework) {
 			ComponentDescriptorName: "github.com/gardener/landscaper/helm-deployer",
 			BlueprintResourceName:   "helm-deployer-blueprint",
 			DeployItem: func(state *envtest.State, target *lsv1alpha1.Target) (*lsv1alpha1.DeployItem, error) {
-				ref := "eu.gcr.io/gardener-project/landscaper/tutorials/charts/ingress-nginx:4.0.18"
-				if !f.RunOnShoot {
-					ref = "eu.gcr.io/gardener-project/landscaper/tutorials/charts/ingress-nginx:v0.1.0"
-				}
+				ref := "eu.gcr.io/gardener-project/landscaper/integration-tests/charts/hello-world:1.0.0"
 				return helm.NewDeployItemBuilder().
 					Key(state.Namespace, "my-di").
 					Target(target.Namespace, target.Name).


test/integration/deployers/blueprints/blueprints_tests.go
```diff
@@ -94,10 +94,7 @@ func DeployerBlueprintTests(f *framework.Framework) {
 			ComponentDescriptorName: "github.com/gardener/landscaper/helm-deployer",
 			BlueprintResourceName:   "helm-deployer-blueprint",
 			DeployItem: func(state *envtest.State, target *lsv1alpha1.Target) (*lsv1alpha1.DeployItem, error) {
-				ref := "eu.gcr.io/gardener-project/landscaper/tutorials/charts/ingress-nginx:4.0.18"
-				if !f.RunOnShoot {
-					ref = "eu.gcr.io/gardener-project/landscaper/tutorials/charts/ingress-nginx:v0.1.0"
-				}
+				ref := "eu.gcr.io/gardener-project/landscaper/examples/charts/hello-world:1.0.1"
```
achimweigel, line 97: I know that Robert is moving these examples sometime to new locations. Perhaps it would be a good idea to upload the chart to this location https://console.cloud.google.com/gcr/images/gardener-project/eu/landscaper/integration-tests/charts for the integration tests. There is already a simple script you could adapt and use for this: https://github.com/gardener/landscaper-examples/blob/master/hello-world/commands/create-and-push-chart.sh

achimweigel (changes requested):

test/integration/deployers/blueprints/blueprints_tests.go
Diaphteiros: Good point.

achimweigel (approved):
/lgtm

PR #681, opened 2023-02-07
make build work on arm / prepare multiarch build

Diaphteiros:

/area dev-productivity
/kind enhancement
/priority 3

`make build-resources` on an arm64 system would produce images which cannot run on amd64 based k8s clusters. This should fix it.

**Release note**:

```other developer
`make build-resources` will now also work on arm64 machines.
```

### DIFF
--- a/.ci/component_descriptor
+++ b/.ci/component_descriptor
@@ -8,12 +8,20 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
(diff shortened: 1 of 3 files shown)


Diaphteiros:
I just tested it, seems to work now.

achimweigel (approved):
/lgtm

PR #705, opened 2023-03-07
fix nil pointer in pickup timeout detection

Diaphteiros:

/area robustness
/kind regression
/priority 3

Split from #669 

**Release note**:

```bugfix operator
Fixed a bug in the pickup timeout detection which would cause the Landscaper to crash once upon creation of a new DeployItem with an empty status.
```


achimweigel (approved):
/lgtm

PR #706, opened 2023-03-07
not finding a resource is not an error while garbage collecting

Diaphteiros:

/area logging
/kind cleanup
/priority 3

Split from #669 

**Release note**:

```other user
The container deployer's garbage collector now doesn't treat not finding a secret or pod as an error anymore. This should significantly reduce logging outputs of the container deployer pod.
```


achimweigel (approved):
/lgtm

PR #704, opened 2023-03-07
Phases Renaming

Diaphteiros:
**How to categorize this PR?**

/area quality
/kind cleanup
/priority 3

**Special notes for your reviewer**:
Split from #669 

**Release note**:

```breaking developer
The phases have been refactored. This includes:
- Executions now use an `ExecutionPhase` and DeployItems use a `DeployItemPhase` and a `DeployerPhase`.
- Values for phases are now accessed via a common struct, e.g. `v1alpha1.InstallationPhases.Failed` instead of `v1alpha1.InstallationPhaseFailed`.
- Phases now have helper methods, e.g. `IsFinal()`.
- DeployItems now use a `InitDelete` and `DeleteFailed` phase when in deletion.
```
```breaking user
When in deletion, DeployItems will now use `InitDelete` and `DeleteFailed` instead of `Init` and `Failed`, respectively.
```

PR #708, opened 2023-03-07
Remove second phase from DeployItem

Diaphteiros:

/area quality
/kind cleanup
/priority 3

Split from #669 

This PR is based on #704 and #711 and requires both to be merged first.

**Release note**:

```breaking user
The `status.deployItemPhase` (in YAML) field in DeployItems has been deprecated and will be removed in the future. It should automatically be removed from all DeployItem resources which are handled by Landscaper. Use `.status.phase` to get information about the current state of the DeployItem instead.
```
```breaking developer
The second phase has been removed from DeployItems (strictly speaking, it has been changed to a pointer which is set to nil during DeployItem defaulting and will be removed in the future).
```


pkg/landscaper/execution/classification.go
```diff
@@ -140,7 +140,7 @@ func newDeployItemClassificationForDelete(executionJobID string, items []*execut
 			if item.DeployItem.Status.GetJobID() != item.DeployItem.Status.JobIDFinished {
 				c.runningItems = append(c.runningItems, item)
 			} else if item.DeployItem.Status.GetJobID() == item.DeployItem.Status.JobIDFinished &&
-				item.DeployItem.Status.DeployerPhase != lsv1alpha1.DeployerPhases.Failed {
+				item.DeployItem.Status.Phase != lsv1alpha1.DeployItemPhases.Failed {
```
robertgraeff, line 143: If we distinguish between phases `Failed` and `DeleteFailed`, then here it should be `DeleteFailed`. We could also use the `IsFailed` method.

robertgraeff (changes requested):

pkg/landscaper/execution/classification.go
Diaphteiros: That's what the `IsFailed` function is for. I just missed that line. Will fix it.

pkg/deployer/container/container_reconcile.go
```diff
@@ -178,44 +178,50 @@ func (c *Container) Reconcile(ctx context.Context, operation container.Operation
...
+		logger.Debug("Deleting pod, as it has finished", "podStatus", pod.Status.Phase)
+		if err := c.CleanupPod(ctx, pod); err != nil {
+			return err
+		}
 	}
-
-	// only remove the finalizer if we get the status of the pod
-	if err := c.CleanupPod(ctx, pod); err != nil {
-		return err
+	if c.ProviderStatus != nil && c.ProviderStatus.PodStatus != nil && c.ProviderStatus.PodStatus.LastSuccessfulJobID != nil && *c.ProviderStatus.PodStatus.LastSuccessfulJobID == c.DeployItem.Status.JobID {
```
achimweigel, line 210: I guess `*c.ProviderStatus.PodStatus.LastSuccessfulJobID == c.DeployItem.Status.JobID` must not be required here. This could not happen otherwise we made a mistake in our model. The reconcile method should only be called if the jod ids differ.

achimweigel: Sorry my mistake - I did not get it right :-)

achimweigel (approved):
/lgtm

robertgraeff (approved):

PR #709, opened 2023-03-08
minor fixes

Diaphteiros:

/area quality
/kind enhancement
/priority 3

This PR adds some minor improvements:
- Fixes a bug in the container deployer where annotations were checked instead of labels.
- Adds a debug log if the container deployer decides that a new pod should be run.
- The `landscaper.gardener.cloud/operation=test-reconcile` annotation now also works if `updateOnChangeOnly` is set to `true` for container deployitems
  - To achieve this, the current reconcile is not aborted anymore after removing the annotation and generating a new job id. This means that there will always be a second reconciliation queued after the current one, but since the test-reconcile annotation is supposed to start a new full reconciliation, which would have changed the deployitem and caused a second reconciliation anyway, this shouldn't have any relevant impact.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/deployer/lib/controller.go
+++ b/pkg/deployer/lib/controller.go
@@ -218,6 +218,7 @@ func (c *controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			}
 
 			// initialize deployitem for reconcile
+			logger.Debug("Setting deployitem to phase 'Init'", "updateOnChangeOnly", di.Spec.UpdateOnChangeOnly, lc.KeyGeneration, di.GetGeneration(), lc.KeyObservedGeneration, di.Status.ObservedGeneration, lc.KeyDeployItemPhase, di.Status.Phase)
 			di.Status.Phase = lsv1alpha1.DeployItemPhases.Init
 			di.Status.DeployerPhase = lsv1alpha1.DeployerPhases.Progressing
 			if err := c.updateDiForNewReconcile(ctx, di); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/deployer/lib/controller.go
```diff
@@ -381,9 +378,10 @@ func (c *controller) removeTestReconcileAnnotation(ctx context.Context, di *lsv1
 	logger, ctx := logging.FromContextOrNew(ctx, []interface{}{lc.KeyReconciledResource, client.ObjectKeyFromObject(di).String()})
 
 	logger.Info("remove test-reconcile annotation")
+	old := di.DeepCopy()
 	delete(di.Annotations, lsv1alpha1.OperationAnnotation)
-	if err := c.Writer().UpdateDeployItem(ctx, read_write_layer.W000149, di); client.IgnoreNotFound(err) != nil {
-		return lserrors.NewWrappedError(err, "RemoveTestReconcileAnnotation", "UpdateDeployItem", err.Error())
+	if err := c.Writer().PatchDeployItem(ctx, read_write_layer.W000149, di, client.MergeFrom(old)); client.IgnoreNotFound(err) != nil {
```
achimweigel, line 383: I would not switch to a patch. One fundamental concept of the overall reconciliation is that we are only working with optimistic locking which is a fundamental concept we should not break anywhere. The logic is too complex to introduce such special stuff again. Especially for a test case.

achimweigel (changes requested):

achimweigel (approved):
/lgtm

PR #718, opened 2023-03-21
Render ImportExecutions in blueprint renderer

Diaphteiros:

/area dev-productivity
/kind enhancement
/priority 3

This PR causes the `landscapercli blueprint render` command to process `ImportExecutions`.

**Release note**:

```feature user
The `landscapercli blueprint render` command will now process the defined `ImportExecutions` (instead of ignoring them).
```
```bugfix user
DeployItem templates without a target reference will not cause a nil pointer exception anymore during the `landscapercli blueprint render` command.
```

### DIFF
--- a/pkg/utils/landscaper/blueprint.go
+++ b/pkg/utils/landscaper/blueprint.go
@@ -7,6 +7,7 @@ package landscaper
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/gardener/landscaper/pkg/utils/dependencies"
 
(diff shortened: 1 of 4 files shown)


pkg/utils/landscaper/blueprint.go
```diff
@@ -177,63 +230,66 @@ func (r *BlueprintRenderer) renderDeployItems(input *ResolvedInstallation, impor
 	// includes resolving target import references to target object references
 	deployItemTemplates := make(core.DeployItemTemplateList, len(executions))
 	for i, elem := range executions {
-		target := &core.ObjectReference{
```
Diaphteiros, line 180: This change looks huge, but I actually just enclosed the coding which is responsible for resolving targets in an `if`, so that it is only executed if a target reference is given (this scenario caused a nil pointer error before).

robertgraeff (approved):

PR #730, opened 2023-04-12
Bugfix: duplicate CDs in list of components

Diaphteiros:

/area robustness
/kind enhancement
/priority 3

When templating in blueprints, it was possible to access the list of all referenced components, including transitively referenced ones. There were a few problems, though:
1. This binding was not mentioned in the documentation.
2. The code which resolves the component references had some bugs which lead to a lot of duplicates in the generated list.

This PR adresses both of these issues.

The solution would be way nicer if Golang had some kind of Set structure which allowed for a custom comparator - we could use a Set of component descriptors which are considered equal if their name and version are equal, respectively. Unfortunately, Golang doesn't have a Set structure at all and the implementation from the k8s apimachinery library does not allow for custom comparators. In theory, it should still work with the default field-wise comparison, but since CDs can become quite big, I implemented this via a map instead, using a struct with name and version as the key.

**Release note**:

```doc user
The `components` binding which is available in the blueprint templating and contains a list of all referenced components is now mentioned in the documentation.
```
```bugfix user
The `components` binding which is available in the blueprint templating will not contain duplicates anymore.
```

### DIFF
--- a/docs/usage/Blueprints.md
+++ b/docs/usage/Blueprints.md
@@ -346,6 +346,11 @@ All template [executions](./Templating.md) get a common standardized binding:
   the component descriptor of the owning component
 
 
+- **`components`**
+
+  the component descriptors of all referenced components, either directly or transitively (includes the component descriptor of the owning component too)
+
+
 - **`blueprintDef`**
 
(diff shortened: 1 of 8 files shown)


achimweigel (approved):
/lgtm

PR #735, opened 2023-04-17
Enable import executions

Diaphteiros:

/area usability
/kind enhancement
/priority 3

Enables the importExecutions which are already described in the documentation, but weren't actually executed in the Landscaper.

**Release note**:

```bugfix user
The blueprint's `importExecutions` will now be executed. The feature was already documented, but the corresponding function was never called.
```
```bugfix user
Fixed a bug in the library which is used for the `blueprint render` subcommand of the landscaper-cli. The bug caused all imports to disappear when importExecutions were used and they were not explicitly rendered into the result, instead of the result of the importExecutions being added to the given imports (without deleting them), as it was intended.
```


pkg/landscaper/installations/imports/constructor.go
```diff
@@ -270,3 +281,41 @@ func (c *Constructor) templateDataMappings(
...
+// Has to be called after import construction (c.Construct(...))
+func (c *Constructor) RenderImportExecutions() error {
+	cond := lsv1alpha1helper.GetOrInitCondition(c.Operation.Inst.GetInstallation().Status.Conditions, lsv1alpha1.ValidateImportsCondition)
+
+	templateStateHandler := template.KubernetesStateHandler{
+		KubeClient: c.Operation.Client(),
+		Inst:       c.Operation.Inst.GetInstallation(),
+	}
+	targetResolver := secretresolver.New(c.Operation.Client())
+	tmpl := template.New(gotemplate.New(c.Operation.BlobResolver, templateStateHandler, targetResolver), spiff.New(templateStateHandler))
```
achimweigel, line 295: Just one question. Why requires a go-templater a blobresolver and a target resolver and a spiff-templater not?

achimweigel (approved):
/lgtm

achimweigel:
Should we perhaps add some integration test for import executions or extend one of the existing ones?

pkg/landscaper/installations/imports/constructor.go
Diaphteiros: I was curious, so I checked:
There seems to be an asymmetry in the available custom functions between gotemplate and spiff:
For gotemplate, there is a `resolve` function, which requires the `blobResolver` and multiple `get*Kubeconfig` functions, which require the `targetResolver`. For spiff, neither of these functions are available, therefore the arguments are not required.

Generally speaking, I think we should keep the custom functions symmetric between all templating implementations. However, since spiff will likely not be used by many people, I don't think that the current state will actually bother anyone. So let's keep it this way until someone complains.

Diaphteiros:
> Should we perhaps add some integration test for import executions or extend one of the existing ones?

I think that would be a good idea, especially since we already missed that they weren't working. I will add some tests.

PR #761, opened 2023-06-14
unblock helm releases stuck in a pending state

reshnm:

/area helm-deployer
/kind enhancement
/priority 3

When a helm deployment (install, upgrade) is running, a helm release secret in a pending state (pending-install, pending-upgrade) is created in the target namespace.
When the helm deployment is interrupted (deployer restart etc.) during the install or upgrade, the helm process will be forever stuck in a pending state until the helm release secret is being deleted.

With this PR the Landscaper helm deployer will try to delete the latest helm release secret that is in a pending state whenever a helm error is encountered.
This unblocks the pending state and retriggers the install, upgrade.

Fixes #

**Release note**:

```feature user
Try to delete pending helm release secret when helm operation was interrupted.
```

### DIFF
--- a/pkg/deployer/helm/helm_suite_test.go
+++ b/pkg/deployer/helm/helm_suite_test.go
@@ -560,7 +560,7 @@ var _ = Describe("Template", func() {
 				},
 			}
 
-			Expect(testenv.Client.Create(ctx, testNamespace))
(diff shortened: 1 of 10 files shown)


achimweigel (approved):
/lgtm

only some tests were not successful

PR #769, opened 2023-06-30
new manual component descriptor logic

reshnm:

/area delivery
/kind cleanup
/priority 3

Since landscaper is still using component-cli, this PR uses its own component upload logic instead of the implicit functionality of the CICD pipeline tools.

The implicit pipeline logic will now upload a dummy component with a version ending with "-dummy".
This is done because the pipeline relies on this component to create a release.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.ci/component_descriptor
+++ b/.ci/component_descriptor
@@ -8,7 +8,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-echo "create dummy component descriptor"
+echo "create dummy component descriptor BEGIN"
 
 SOURCE_PATH="$(dirname $0)/.."
 VERSION="$(${SOURCE_PATH}/hack/get-version.sh)"
@@ -17,6 +17,10 @@ COMPONENT_DESCRIPTOR_DIR="${SOURCE_PATH}/../${COMPONENT_DESCRIPTOR_DIR}"
 
 VERSION="${VERSION}-dummy" REPO_CTX="${REPO_CTX}" COMPONENT_DESCRIPTOR_DIR="${COMPONENT_DESCRIPTOR_DIR}" python ${SOURCE_PATH}/hack/create-dummy-cd.py
 
-echo "create real component descriptor"
+echo "create dummy component descriptor END"
+
+echo "create real component descriptor BEGIN"
 
 ${SOURCE_PATH}/.ci/component_descriptor_real
+
+echo "create real component descriptor END"
--- a/.ci/component_descriptor
+++ b/.ci/component_descriptor
@@ -13,7 +13,21 @@ echo "component descriptor dummy step"
 VERSION="$(${SOURCE_PATH}/hack/get-version.sh)"
 REPO_CTX="eu.gcr.io/gardener-project/development"
 
-${COMP_CLI}  ca create "${COMPONENT_DESCRIPTOR_DIR}" \
-  --component-name=github.com/gardener/landscaper \
-  --component-version=${VERSION} \
-  --repo-ctx=${REPO_CTX}
+COMPONENT_DESCRIPTOR="
+meta:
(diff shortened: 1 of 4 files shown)


hack/create-dummy-cd.py
```diff
@@ -0,0 +1,17 @@
...
+repo_ctx = os.environ["REPO_CTX"]
+component_descriptor_dir = os.environ["COMPONENT_DESCRIPTOR_DIR"]
+
+cd = None
+
+with open(os.path.join(component_descriptor_dir, "base_component_descriptor_v2"), "r") as base_cd_file:
+    cd = yaml.safe_load(base_cd_file.read())
+
+cd["component"]["version"] = version
+cd["component"]["repositoryContexts"][0]["baseUrl"] = repo_ctx
```
AndreasBurger, line 14: Not needed, the pre-generated CD should already have these values set.

.ci/component_descriptor
```diff
@@ -1,95 +1,26 @@
...
+VERSION="${VERSION}-dummy" REPO_CTX="${REPO_CTX}" COMPONENT_DESCRIPTOR_DIR="${COMPONENT_DESCRIPTOR_DIR}" python ${SOURCE_PATH}/hack/create-dummy-cd.py
 
-printf "> add component references\n"
-${COMP_CLI} ca component-references add ${LS_CA_PATH} \
-    VERSION=${VERSION} \
-    ${SOURCE_PATH}/.landscaper/component-references.yaml
+echo "create dummy component descriptor END"
 
-cat ${LS_CA_PATH}/component-descriptor.yaml
+echo "create real component descriptor BEGIN"
```
AndreasBurger, line 22: I suppose the point here is to create _and push_ it so that some other stuff downstream (e.g. integration-tests) can rely on it. Might be useful to log that as well, maybe like
```bash
echo "Creating real component descriptor and uploading it"
```

.ci/component_descriptor
```diff
@@ -1,95 +1,26 @@
...
-buildComponentArchive "container-deployer"
-buildComponentArchive "helm-deployer"
-buildComponentArchive "manifest-deployer"
-buildComponentArchive "mock-deployer"
-
-# add landscaper component descriptor
-printf "> Create Landscaper ca archive\n"
-LS_CA_PATH="$(mktemp -d)"
-cp ${BASE_DEFINITION_PATH} "${LS_CA_PATH}/component-descriptor.yaml"
+REPO_CTX="eu.gcr.io/gardener-project/development"
```
AndreasBurger, line 15: ideally this shouldn't be hardcoded.
The base-url is currently passed along as `CURRENT_COMPONENT_REPOSITORY` in the component-descriptor script, but if you plan to reuse that later that's also not really an option.

.ci/component_descriptor
reshnm: There is some log outputs later that log the publishing of the components.

.ci/component_descriptor
reshnm: I rather not rely on implicit environment variables when this script is meant to run indipendently (and derive from standard CI/CD logic) from the CI/CD versions. The dev repo was already hardcoded for a long time.

In the future I would rather like to discuss how CI/CD tooling can support OCM transport formats created with OCM cli instead of expecting a plain component descriptor yaml file.

hack/create-dummy-cd.py
reshnm: These values are actually overwritten so that they don't interfere with the component descriptor we push manually.
We are not using this component descriptor. We use it only to satisfy the CICD requirements.
The repository context will replaced with the dev repo and the version will get a "-dummy" postfix.

achimweigel (approved):
/lgtm

PR #770, opened 2023-07-04
Revert "Horizontal scaling of deployer "

robertgraeff:
Reverts gardener/landscaper#753

The scaling of deployers should not be part of the current release. Therefore we revert that pull request.


reshnm (approved):
/lgtm

PR #775, opened 2023-07-06
Skip readiness check for disabled subcharts

### LINKED ISSUE #774: Installation of chart with subcharts is not handled as expected

/area helm-deployer
/kind bug
/priority 3

**What happened**:
When installing a chart which has subcharts the landscaper manages to install the main chart successfully, however when the subchart is "disabled" (it should not be installed) it looks like it still tries to render/create (from helm perspective) the resources of the subchart.

I expected only the main chart to be installed and the subchart to not be involved at all.

I have not tried to reproduce the issue with different helm chart from the one of Connectivity Service. It would be hard to navigate you how to install our chart and it would require subaccounts in CF and so on. I am available to help reproduce the issue at any time. Also I can leave a failed installation in the landscaper instance if needed for troubleshooting. Or I guess you could update the installation resource I used and reproduce it.

I am attaching the chart which I am trying to install. 
[connectivity-proxy-k8s-helm-2.9.0.tgz](https://github.com/gardener/landscaper/files/11964943/connectivity-proxy-k8s-helm-2.9.0.tgz)

If I try to install the chart with the subchart enabled it works fine.

**Environment**:

- Landscaper version: v0.40.0
- Kubernetes version (use `kubectl version`): v1.25
- Others:

robertgraeff:

/area helm-deployer
/kind bug
/priority 3

Suppose you use the "real" helm deployer to deploy a helm chart with a subchart. Suppose the subchart is disabled, i.e. the `Chart.yaml` file (resp. `requirements.yaml`) contains a condition whose value is false:

```yaml
dependencies:
  - name: my-subchart
    condition: subchart-enabled  # path into the helm values that evaluates to false
    repository: ...
    version: ...
``` 

In this situation the helm deployment of the subchart was skipped, which is correct. However, the default readiness check afterwards did not ignore the resources of the subchart, which is wrong. 

For example, if the subchart contained a `Deployment` resource, then this `Deployment` was not deployed on the target cluster, but the readiness check nevertheless waited for it to become ready. As a result, the corresponding `DeployItem` would fail with a timeout of the readiness check.

```yaml
  lastError:
    codes:
    - ERR_READINESS_CHECK_TIMEOUT
    message: 'recoverable error: unable to get apps/v1, Kind=Deployment ........ not found'
    operation: DefaultCheckResourcesReadinessHelm
    reason: CheckResourceReadiness
```

Fixes #774 

A remark about the implementation: the helm deployers have a list of deployed resources in their status. The "real" helm deployer needs this list only for the default readiness check (whereas the manifest-helm deployer needs the list also for the delete.) In case of the real helm deployer, we will now obtain the list by using helm's get release method (`action.NewGet`). In this way, we get only the actually deployed resources, i.e. not the resources of disabled subcharts. Moreover, we filter this list, and write only those resources to the status which are needed for the default readiness check (Deployments, ReplicaSets, etc.)    

**Release note**:

```bugfix user
- bugfix in the helm deployer: skip readiness check for resources of disabled subcharts
```

### DIFF
--- a/pkg/deployer/helm/ensure.go
+++ b/pkg/deployer/helm/ensure.go
@@ -60,11 +60,6 @@ func (h *Helm) ApplyFiles(ctx context.Context, files, crds map[string]string, ex
 		}
 	}
 
-	manifests, err := h.createManifests(ctx, currOp, files, crds)
-	if err != nil {
-		return err
-	}
-
 	var (
 		managedResourceStatusList managedresource.ManagedResourceStatusList
 		deployErr                 error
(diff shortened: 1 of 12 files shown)


achimweigel (approved):
/lgtm

PR #802, opened 2023-08-09
Enhance controller-utils library

Diaphteiros:
**How to categorize this PR?**

/area dev-productivity
/kind enhancement
/priority 3

**What this PR does / why we need it**:
Adds some fixes and new code to the controller-utils library:
- `logging.StartReconcileFromContext` now also returns a new context containing the enriched logger. This is a breaking change, but with little consequences and easy mitigation.
- An `errors` package has been added to the library which enables aggregating multiple errors into one.
- Fixed a bug in the certificate library which would hide the actual error when creating a cert secret failed. Also fixed a duplicate logging key there.
- Added a `HasOwnerReference` function.
- Moved most of the code which created and configured the landscaper webhooks server into a new webhook library. It should be generic enough to be used for other projects too.

**Additional information for reviewers**:
For how the webhook library reduces code when creating a webhook server, see these changes:
- https://github.com/gardener/landscaper/pull/803/files#diff-bddaa641dcdcefb4adb5b95e2f893bb8c4fcc537cc302ee7486d42bd0ce9b31c
- https://github.com/gardener/landscaper/pull/803/files#diff-077e02833b7f142a121fefde1547df40680d79c437cf44c1a4c5ace128c5fd57
- https://github.com/gardener/landscaper/pull/803/files#diff-38286498d9e6854874aa29f969b5c5a030b693ef1814263aa17f6fa22b08b839
- https://github.com/gardener/landscaper/pull/803/files#diff-146bc4c2f5244b75416f9521fd022b37aa710bc96ebe4ab6dd1a6985364fd8c4
- https://github.com/gardener/landscaper/pull/803/files#diff-b5ac2b02cf919bb8f68b90a7dd4542189b357fcf13ea2796a0026f9a9c0e436c

**Release note**:

```breaking developer
`logging.StartReconcileFromContext` now also returns a new context containing the enriched logger.
```
```breaking developer
The signature of the controller-utils library's `GenerateCertificates` function has changed: Instead of only returning the CA bundle, it now returns the CA and server certificates. For migration, replace `caBundle, err := GenerateCertificates(...)` with `caCert, _, err := GenerateCertificates(...); caBundle := caCert.CertificatePEM`.
```
```other developer
An `errors` package has been added to the library which enables aggregating multiple errors into one.
```
```bugfix developer
Fixed a bug in the certificate library which would hide the actual error when creating a cert secret failed. Also removed a duplicate logging key.
```
```other developer
[...]

PR #803, opened 2023-08-09
Use new webhook library for landscaper webhooks server

Diaphteiros:

/area quality
/kind cleanup
/priority 3

Replaces the implementation of the landscaper webhooks server, so that it now uses the new webhook library. Most of the coding has been moved into the library, so this significantly reduces the amount of code in the landscaper that is required to run the webhook server.

#802 needs to be merged before this can be merged.
Due to the webhook-relevant CLI flags now being provided by the library, the flag for the certificate secret namespace has changed from `--certificates-namespace` to `--cert-ns`. This is a breaking change. The webhook server's helm chart has been adapted accordingly.

**Release note**:

```other developer
The landscaper now uses the new webhook library to configure the landscaper webhooks server.
```
```breaking operator
Due to switching to the webhook library, the `--certificates-namespace` flag, which could be used to set the namespace of the certificate secret for the landscaper webhook server, has changed to `--cert-ns`. The landscaper helm chart has been adjusted accordingly.
```


Diaphteiros:
/hold
Don't merge before #802 is merged.

achimweigel (approved):
/lgtm

PR #808, opened 2023-08-31
enforce KeyResource values to be of type string

reshnm:

/area logging
/kind bug
/priority 3

Logging parameter types must be consistent. There have been some logs where the `KeyResource` value was not of type string.

Fixes #

**Release note**:

```bugfix operator
enforce KeyResource values to be of type string
```


achimweigel (approved):
/lgtm

PR #809, opened 2023-08-31
Auto Locking

robertgraeff:

/area scalability
/kind task
/priority 3

#### (A) Enabling locking based on the HPA configuration

The Landscaper is prepared for horizontal scaling of the main controllers (Installation and Execution controller) and the deployers since pull request #784. The maximum number of pods for these controllers can be configured in the values of the corresponding helm chart. Additionally, a feature flag `LockerEnabled` (in `pkg/utils/lock/locker.go`) needed to be set to `true`, because controllers running on more than one pod must lock the resources during reconciliation. 

With this pull request we get rid of the feature flag `LockerEnabled`. Instead, locking is automatically enabled if the maximum number of replicas is > 1.

#### (B) Deploy strategy

During an update, old terminating pods and new pods might temporarily run in parallel. If the configured maximum number of pods is > 1, locking is enabled, so that this is not a problem. However, if the maximum number of pods for a controller is 1, locking is not enabled. In this case, we want to stop the old pod before the new one is started. Therefore, we set the  [deployment strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy) to `Recreate` instead of `RollingUpdate`. 

#### (C) Checksum of the deployer configuration

At the pods of the manifest deployer, an annotation was missing that contains a hash of the deployer configuration. The annotation ensures that a new pod is started if the configuration was changed. This pull request adds this annotation. The other deployers already have such an annotation.

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel (approved):
/lgtm

PR #822, opened 2023-09-20
Component Library Flag

robertgraeff:

/area oci
/kind enhancement
/priority 3

This pull request extends the Installation, Execution, and DeployItem spec with a field `ComponentLibrary`. At a root Installation, a user can set this field to specify whether the `cnudie` or `ocm` component library should be used during the reconciliation. The default is currently `cnudie`. Each object passes the field to its subobjects. 

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/apis/core/v1alpha1/types_shared.go
+++ b/apis/core/v1alpha1/types_shared.go
@@ -369,6 +369,7 @@ func (r VersionedResourceReference) ObjectMeta() cdv2.ObjectMeta {
 type ComponentLibrary string
 
 const (
-	ComponentLibraryCNUDIE ComponentLibrary = "cnudie"
-	ComponentLibraryOCM    ComponentLibrary = "ocm"
+	ComponentLibraryCNUDIE  ComponentLibrary = "cnudie"
+	ComponentLibraryOCM     ComponentLibrary = "ocm"
+	ComponentLibraryDefault ComponentLibrary = ComponentLibraryCNUDIE
 )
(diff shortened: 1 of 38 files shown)


apis/core/v1alpha1/types_shared.go
```diff
@@ -365,3 +365,11 @@ func (r VersionedResourceReference) ObjectMeta() cdv2.ObjectMeta {
...
 	}
 }
+
+type ComponentLibrary string
+
+const (
+	ComponentLibraryCNUDIE  ComponentLibrary = "cnudie"
+	ComponentLibraryOCM     ComponentLibrary = "ocm"
+	componentLibraryDefault ComponentLibrary = ComponentLibraryCNUDIE
+)
```
achimweigel, line 375: Why are there 3 constants and not two compared to the core types?

achimweigel (changes requested):

apis/core/v1alpha1/types_shared.go
robertgraeff: It defines which of the two is the default.

At some point in the future we wanted to switch the default from `cnudie` to `ocm`. That will be done here, at a single point, by switching from 
```golang
componentLibraryDefault ComponentLibrary = ComponentLibraryCNUDIE
```
to
```golang
componentLibraryDefault ComponentLibrary = ComponentLibraryOCM
```
instead of inspecting all places where the `cnudie` constant occurs.

achimweigel: Oh yes, I totally missed this.

achimweigel (approved):
/lgtm

PR #823, opened 2023-09-21
Proposal timeouts and typical error page

achimweigel:

/area deployers
/kind enhancement
/priority 3

This PR contains a proposal for timeouts and add a page where to list typical errors when working with Landscaper.

Fixes #

**Release note**:

```other operator
NONE
```


docs/proposals/timeouts.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Solution proposal
+
+- Remove pickup timeout, set it on a very large value or add some DeployItem specific setting:
+  - The main reason for the pickup timeout was to resolve the problem, that if someone wrongly configures a DeployItem
+    such that no deployer would be responsible, that this DeployItem is finished. This use case is mainly relevant during
+    development. Comparing this with the problem of failed DeployItems in productive systems under heavy load, seems to
+    be a bad deal. Furthermore, for the development scenario, it is already possible to find out that a DeployItem
+    is not picked up, by its empty status and such DeployItems could be stopped by the interrupt annotation. 
+    Therefore, the removal of this timeout or setting it on a high value might be a good idea.
```
Diaphteiros, line 85: I disagree here. Besides showing errors during development, the pickup timeout has a second purpose: it detects if the corresponding deployer is not running. Imagine the following scenario: You are using a managed Landscaper service, so you don't have access to the deployer pods. The deployer responsible for your DeployItem was running, so the DI has a status, but then it crashed for whatever reason and is unable to come up again. Now you apply some change to your DI. This change will never get reconciled and to recognize the indicators of the deployer not running you would need detailed knowledge about the Landscaper-internal logic (and even then, most of these indicators could also hint to other failures).
The result: Your changes are not applied and you have no way to find out why.

For this reason, I would strongly advise against removing the pickup timeout. It should already be possible to deactivate it via the Landscaper configuration. 
I don't have a strong opinion about the duration - of course it would be nice to detect configuration errors and missing deployers as fast as possible, but running into pickup timeouts when the system is under load is pretty annoying and should be prevented. Maybe half an hour or an hour would be a good duration - if we hit that under load, it's a strong sign we need to scale up.

docs/proposals/timeouts.md
```diff
@@ -0,0 +1,119 @@
...
+    be a bad deal. Furthermore, for the development scenario, it is already possible to find out that a DeployItem
+    is not picked up, by its empty status and such DeployItems could be stopped by the interrupt annotation. 
+    Therefore, the removal of this timeout or setting it on a high value might be a good idea. 
+  - If we want to be more defencive, we could add a pickup timeout to the DeployItem specification itself. Then the 
+    customer could decide by itself if he wants this feature e.g. in his development scenario.
+
+- Reduce number of processing timeout settings
+  - Motivation: It is less complicated for a user to only set one timeout for the deployment or deletion of a DeployItem.
+    Even finding here the right value is already complicated but to take all the other current settings into account
+    is nearly impossible.
```
Diaphteiros, line 92: I completely agree with that. Even if we have fine-granular timeout settings (which likely nobody will use), there should be an easy way to increase all of them at once for slow components.

docs/proposals/timeouts.md
```diff
@@ -0,0 +1,119 @@
...
+  - Only one processing timeout on deploy item level which determines how long the deployment is allowed to require.
+  - Only one delete timeout on deploy item level which determines how long the deletion is allowed to require.
+    (if not set, use some default or use same as for progressing timeout)
+  - Timeout controller:
+    - Exclude the manifest and helm deployer from the timeout controller
+    - When we have also found a similar solution for the container deployer we could perhaps completely remove the central
+      processing timeout controller?
```
Diaphteiros, line 99: While I think that's a good idea to move the progressing timeout logic into the deployers, removing the central logic has the disadvantage that a progressing DeployItem whose deployer dies (and fails to come up again) will be stuck in `Progressing` for a long time. Although this case is probably rare enough to not be worth the problems with the central timeout you mentioned.

docs/proposals/timeouts.md
```diff
@@ -0,0 +1,119 @@
...
+      - Running requests like helm 3 deployments or manifest deployments are canceled if they require too much time
+      - The helm 3 timeout is recomputed with respect to the remaining duration - no matter if it is set or not.
+    - Deletions are handled the same way. 
+
+## Open Question
+
+- When does the processing timeout start? When a DeployItem is triggered of when the deployer starts working on it?
+  - Alternative one has the advantage, that a system under load, where it might need some time until a deployer could
+    pick up a DeployItem, is not running out of time. It seems also to be much easier for a customer to know how long a 
+    deployment for a DeployItems needs, compared to how long a deployer requires to pick up a DeployItem.
```
Diaphteiros, line 113: I'm confused - did you mean alternative two?

docs/proposals/timeouts.md
```diff
@@ -0,0 +1,119 @@
...
+  
+- In case of an error and an immediate retry of the reconciliation of the DeployItem, should the required time of the 
+  erroneous reconciliations from before be taken into account and removed for the allowed processing time? 
+  Probably yes as it prevents infinite loops.
```
Diaphteiros, line 117: Hm, but the controller-runtime applies some exponential backoff logic for repeatedly failed reconciliations. Imagine a DI which talks to a server and that server is down for five minutes. When the server is back up and the repeatedly failed DI is reconciled again, it immediately runs into a progressing timeout, because the past five minutes of trying count into the timeout duration? Doesn't sound like a good idea to me ...

I don't think that repeated failures of a DI are a problem, because the DI will be in Error state for most of the time, so it's easy to recognize. But maybe I'm missing something here.

Diaphteiros (commented):
I just dumped some thoughts of mine here, so I don't forget them until the discussion.

docs/proposals/timeouts.md
achimweigel: Is there really such a big difference for a user with no access to the deployer, if the DI remains in a state with a status.transitionTime.initTime not set or that he sees that a pickup timeout has occurred. In both cases the user does only know that no deployer processed the DI so far. Programming against both in e.g. a pipeline, seems to have the same complexity. 

Monitoring that all deployers are running is the responsibility of the health checks which also have the advantage, that they recognise a failing deployer even if no DI is currently processed.

docs/proposals/timeouts.md
achimweigel: If the deployer fails and some DI remains in progressing this seems to be the smallest problem in such a situation. As long as the deployer is not up and running again, the state of DIs does not matter. 

Is it really right, that the contract becomes more complex. From my point of view the current deployer just do not fulfil a reasonable timeout-contract in combination with the timeout controller. The DI is set on failed but the deployer is happily continuing with its work. Several customers were quite surprised that helm charts were successfully deployed even if the DI failed with a processing timeout.

docs/proposals/timeouts.md
achimweigel: Your are totally right - I mean alternative two.

robertgraeff (approved):

PR #830, opened 2023-09-25
update pipeline definition for head updates

fabianburth:
**How to categorize this PR?**

/area TODO
/kind TODO
/priority 3

**What this PR does / why we need it**:
Follow up for #829 

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator

```

PR #836, opened 2023-09-28
Pipeline fix

fabianburth:
If you set an `output_dir` in a step, the pipeline coding implicitly creates a corresponding environment variable with a `_PATH` suffix (e.g. for `output_dir: 'integration_test'` `INTEGRATION_TEST_PATH`. 
Since the `output_dir` was removed in #830 but the integration_test script tried to use this environment variable, the script and therefore, the pipeline, failed.

The integration_test script uses this path to store a file with the integration test logs. This file is then accessed by the update_release.py script to upload the logs as asset of the release. 
The integration test script has been adjusted so that they should write to different `output_dir`s and the update_release.py script adds both log outputs as assets.

/area delivery
/kind bug
/priority 1

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -77,6 +77,7 @@ landscaper:
           trait_depends:
           - publish
           image: 'eu.gcr.io/gardener-project/landscaper-service/integration-test:1.20.6-alpine3.18'
+          output_dir: 'integration_test_ocm'
         integration_test_ocm:
           execute:
           - integration-test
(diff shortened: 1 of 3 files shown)


morri-son (approved):

.ci/pipeline_definitions
```diff
@@ -77,13 +77,15 @@ landscaper:
           trait_depends:
           - publish
           image: 'eu.gcr.io/gardener-project/landscaper-service/integration-test:1.20.6-alpine3.18'
+          output_dir: 'integration_test_ocm'
```
Diaphteiros, line 80:
```suggestion
          output_dir: 'integration_test_cnudie'
```

Diaphteiros (changes requested):
Small copy-paste error, otherwise lgtm.

.ci/pipeline_definitions
fabianburth: Good catch, thanks!

Diaphteiros (approved):
/lgtm

PR #755, opened 2023-05-31
Component model (part 2)

robertgraeff:
**How to categorize this PR?**

/area oci
/kind task
/priority 3

**What this PR does / why we need it**:

This pull request is a continuation of PR #722. 

- It moves the code for the creation of component resolvers into the facade (`pkg/components/cnudie/componentresolvers`)
- It adds errors to the methods of classes `ComponentVersion` and `Resource` of the component model.
- It adds a new factory method for `RegistryAccess` that will be used by the landscaper-cli.  
- It moves the `utils/tar` package into the `component/model` package in order to make the facade self-contained.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```feature developer
- extension of the facade that encapsulates the accesses to OCI registries and Helm chart repositories
```

PR #844, opened 2023-10-04
intergrate active ocm-lib coding

fabianburth:

/area oci
/kind task
/priority 3

This PR is based on #755. 

It provides adapters to implement the introduced facade for working with the ocm with the ocm-lib. Ultimately, this shall enable the landscaper to also work with ocm component descriptor format instead of the legacy cnudie component descriptor format. 

Fixes #

**Release note**:

```feature developer
- support for ocm component descriptors
```


pkg/landscaper/blueprints/resolve.go
```diff
@@ -124,18 +131,25 @@ func Resolve(ctx context.Context, registryAccess model.RegistryAccess, cdRef *ls
...
-	return ResolveBlueprintFromComponentVersion(ctx, componentVersion, bpDef.Reference.ResourceName)
+	resource, err := componentVersion.GetResource(bpDef.Reference.ResourceName, nil)
+	if err != nil {
+		return nil, err
+	}
+	content, err := resource.GetTypedContent(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return content.Resource.(*Blueprint), nil
```
achimweigel, line 142: A cast without a check if critical as it will cause a panic stopping the whole system.

pkg/landscaper/installations/executions/template/template.go
```diff
@@ -288,10 +288,7 @@ func serializeComponentDescriptor(componentVersion model.ComponentVersion) (inte
 		return nil, nil
 	}
 
-	cd, err := componentVersion.GetComponentDescriptor()
-	if err != nil {
-		return nil, err
-	}
+	cd := componentVersion.GetComponentDescriptor()
```
achimweigel, line 291: could this return nil?

pkg/landscaper/jsonschema/testreg/cert.go
```diff
@@ -125,7 +125,7 @@ func GenerateCertificates() (*Certificate, error) {
 const chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
 func init() {
-	mathrand.Seed(time.Now().Unix())
+	mathrand.New(mathrand.NewSource(time.Now().Unix()))
```
achimweigel, line 128: Why this change? I do not understand the construct at all. The generated Rand object is never used.

achimweigel (changes requested):

pkg/landscaper/jsonschema/testreg/cert.go
fabianburth: Linter complained after a dependency upgrade that `Seed()` is deprecated. Replaced it with the suggestion provided within the library, but you are right, it is never used. Removed it.

pkg/landscaper/blueprints/resolve.go
fabianburth: I think we can be certain that we have a blueprint at this point. But you are right, better safe than sorry! Added checks to each occurrence.

pkg/components/registries/factory.go
```diff
@@ -1,101 +1,52 @@
...
+	}
 }
 
-func (f *Factory) NewHelmOCIResource(ctx context.Context,
-	ociImageRef string,
-	registryPullSecrets []corev1.Secret,
-	ociConfig *config.OCIConfiguration,
-	sharedCache cache.Cache) (model.Resource, error) {
-	return f.cnudieFactory.NewHelmOCIResource(ctx, ociImageRef, registryPullSecrets, ociConfig, sharedCache)
+func GetFactory() model.Factory {
```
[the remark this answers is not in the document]
fabianburth: Adjusted. Please check again.

pkg/deployer/helm/chartresolver/chartresolver_suite_test.go
```diff
@@ -61,51 +56,51 @@ var _ = Describe("GetChart", func() {
...
-
-		chartAccess := &helmv1alpha1.Chart{
-			FromResource: ref,
-		}
-
-		chart, err := chartresolver.GetChart(ctx, chartAccess, nil, nil, nil, nil, nil)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(chart.Metadata.Name).To(Equal("ingress-nginx"))
-	})
+	// The team collectively decided that this option to retrieve helm charts is no longer supported
```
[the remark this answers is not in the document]
fabianburth: done

achimweigel (approved):

PR #850, opened 2023-10-09
integration-tests with kubernetes version 1.27

guewa:

/area security
/kind test
/priority 2

Integrtaion tests will run in a cluster with kubernetes version >1.27 where no static kubeconfigs are allowed anymore

Fixes #

**Release note**:

```other operator
NONE
```


achimweigel:
Perhaps we should discuss if we should already upgrade the test cluster to 1.27 as long as the Landscaper clusters in our productive systems are still running on 1.25? Another question is, if we should wait until steampunk is on Canary to upgrade our cluster versions in a coordinated way?

PR #857, opened 2023-10-12
move logic which TargetResolver to use into library

Diaphteiros:

/area dev-productivity
/kind enhancement
/priority 3

Before this change, we had one `TargetResolver` implementation, which can resolve secret references and should be used if there is a secret reference within the Target. The problem with this approach is that someone from the outside who wants to work with Targets (e.g. when using the results (=exports) of an Installation programmatically) would need to know which TargetResolver implementations exist and when to use which one. The only way to figure this out is by checking usages of the TargetResolver within the Landscaper code.
Since it should be as easy as possible to build upon the exports of an Installation, this logic was moved into the `targetresolver` package:
1. There is now a `GenericResolver`. It implements the `TargetResolver` interface, but doesn't resolve anything by its own, instead it checks which other implementation to use and calls that one.
2. The `targetresolver` package now has a `Resolve` function that can resolve any Target. It does so by constructing a GenericResolver internally.

While this makes using TargetResolvers significantly easier, there is a disadvantage: Both methods require all arguments that any `TargetResolver` implementation takes. Currently, there is only one implementation and that only requires a `client.Client` to access the cluster, but if we add another implementation, e.g. for resolving references to a secret vault, both - the GenericResolver as well as the Resolve function - need to be given the client as well as the vault configuration. They could still be instantiated with a nil vault configuration, but then trying to resolve a secret that references a vault would cause an error. This also means that each new `TargetResolver` implementation causes a breaking change (for developers) and requires each call to GenericResolver or the Resolve function to be modified.

However, there are two reasons why this is probably not as bad as it sounds:
1. We used the TargetResolver for quite some time now and did not require any other implementation. It is unlikely that new implementations will be created often in the future.
2. Whenever a Target is resolved, we usually want it to be resolvable by all supported implementations - I can't imagine a place in the code where it would be fine to be able to resolve just some of the allowed Targets and not all of them (except for tests). This means that wherever Targets are resolved, we would require all arguments for all implementations anyway, so having them all gathered in a single function call is not really a drawback.

**Release note**:

```breaking developer
The `NewResolvedTarget` constructor has moved from the `targetresolver` package to the `v1alpha1` api package. This has been done to avoid import cycles in the refactored `targetresolver` package, but it also makes sense, as the `ResolvedTarget` type is also declared there.
```
```feature developer
New options for resolving a `Target` have been added to the `targetresolver` package:
1. The package itself now contains a `Resolve` function.
2. There is a new `generic` sub-package containing the `GenericResolver` implementation of the `TargetResolver` interface.
Both of these have the advantage that they contain the logic when to use which TargetResolver implementation (although there currently is only one, but they will be expanded when new implementations are added). It is strongly recommended to use one of these methods instead of using the SecretRefResolver explicitly.
```

### DIFF
--- a/apis/core/v1alpha1/types_target.go
+++ b/apis/core/v1alpha1/types_target.go
@@ -125,3 +125,15 @@ type ResolvedTarget struct {
 	// Otherwise, the inline configuration of the target is put here.
 	Content string `json:"content"`
 }
+
+// NewResolvedTarget is a constructor for ResolvedTarget.
+// It puts the target's inline configuration into the Content field, if the target doesn't contain a secret reference.
+func NewResolvedTarget(target *Target) *ResolvedTarget {
+	res := &ResolvedTarget{
+		Target: target,
+	}
+	if target.Spec.SecretRef == nil && target.Spec.Configuration != nil {
+		res.Content = string(target.Spec.Configuration.RawMessage)
+	}
+	return res
+}
--- a/controller-utils/pkg/landscaper/targetresolver/generic/genericresolver.go
+++ b/controller-utils/pkg/landscaper/targetresolver/generic/genericresolver.go
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package generic
+
+import (
+	"context"
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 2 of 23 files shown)


achimweigel (approved):
I guess that programming purists will run in circles screaming but I really like this approach. Perhaps, when we add some other target types we do not change the constructor but perhaps add just another one, e.g. NewForVault. This would not break any customer coding.

achimweigel:
Perhaps when implementing something in controller-utils, we could add unit tests in the main pkg folder. Ok, these tests are not nearby the coding but this is anyhow better than no tests and this is the way to go in most other languages. Perhaps we could add a folder under pkg named contoller-util-testing or something like that.

Issue #875, opened 2023-10-31
DeployItems keep being reconciled for more than 6 months


morri-son:
/area helm-deployer
/bug bug|
/priority 3 

**What happened**:
In the logging stack for LIVE there's a repeating issue with the reconciliation of all three deploy items of an installation, e.g. https://dashboards-sf-4b9f9729-2334-43cb-aa27-aa72888cf7b2.cls-01.cloud.logs.services.eu10.hana.ondemand.com/app/dashboards#/view/271d3fc0-f86f-11ed-97fa-61b53b28387e?_g=h@6dc552e&_a=h@731558efor

It is for tenant id k5f74a4p in resource cluster t-k5f74a4p/dev-landscaper-685gz, namespace cu-landscaper-resources-cf-us10-staging. The installation is "ingress-gateway-installation". Its three deployItems and the installation are all succeeded in April 2023, but the Helm deployer in the target landscaper keeps reconciling the DeployItems and fails with 

```
{"level":"error","ts":"2023-10-31T13:12:30.659Z","logger":"deployer.helm","msg":"unable to get landscaper context: contexts.landscaper.gardener.cloud \"context\" not found","reconciledResourceKind":"DeployItem","deployItemType":"landscaper.gardener.cloud/helm","reconcileID":"5c3e7fda-0fd7-4163-8814-74c1409b3fe3","reconciledResource":"cu-landscaper-resources-cf-us10-staging/ingress-gateway-installation-istio-base-mtm79","method":"LogErrorAndGetReconcileResult","error":"unable to get landscaper context: contexts.landscaper.gardener.cloud \"context\" not found"}
```

as the installation refers to a context that has been deleted in the meantime

**What you expected to happen**:
The Installation / DeployItems should be updated to "failed", so that the HelmDeployer stops reconciling.

**How to reproduce it (as minimally and precisely as possible)**:

**Environment**:
LIVE

- Landscaper version: v0.79.0
- Kubernetes version (use `kubectl version`): v1.25.12

achimweigel:
Fixed with PR https://github.com/gardener/landscaper/pull/877

PR #877, opened 2023-11-03
Fix for missing context of deploy items

achimweigel:

/area deployers
/kind bug
/priority 3

This PR fixes the problem described here: https://github.com/gardener/landscaper/issues/875

https://github.com/gardener/landscaper/issues/875

**Release note**:

```other operator
- [Fix for missing context](https://github.com/gardener/landscaper/issues/875)
```

### DIFF
--- a/pkg/deployer/lib/controller.go
+++ b/pkg/deployer/lib/controller.go
@@ -230,12 +230,6 @@ func (c *controller) reconcilePrivate(ctx context.Context, metadata *metav1.Part
 
 	old := di.DeepCopy()
 
-	lsCtx := &lsv1alpha1.Context{}
-	// todo: check for real repository context. Maybe overwritten by installation.
-	if err := c.lsClient.Get(ctx, kutil.ObjectKey(di.Spec.Context, di.Namespace), lsCtx); err != nil {
-		return lsutil.LogHelper{}.LogStandardErrorAndGetReconcileResult(ctx, fmt.Errorf("unable to get landscaper context: %w", err))
-	}
-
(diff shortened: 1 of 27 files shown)


pkg/deployer/lib/controller.go
```diff
@@ -285,22 +279,22 @@ func (c *controller) reconcilePrivate(ctx context.Context, metadata *metav1.Part
 	// Deployitem has been initialized, proceed with reconcile/delete
 
 	if di.DeletionTimestamp.IsZero() {
-		lsError := c.reconcile(ctx, lsCtx, di, rt)
-		err := c.handleReconcileResult(ctx, lsError, old, di)
-		return c.buildResult(di.Status.Phase, err)
+		lsError := c.reconcile(ctx, di, rt)
+		_ = c.handleReconcileResult(ctx, lsError, old, di)
```
Diaphteiros, line 283: I don't get this change. This function returns an error, if updating the status of the DeployItem failed. We want this error to be returned, so that the controller-runtime requeues the DeployItem so that it gets the correct status, don't we? Why would we ignore this error?

achimweigel: In the next line `return c.buildResult(di.Status.Phase)` the re-queue logic is implemented. It only depends on the status of the deploy item and not the error. It is long ago when Robert changed this. As far as I remember we wanted to prevent timeouts due to exponential back-offs. In this change I have just cleaned up a  not used parameter in the method `c.buildResult(...)`

robertgraeff (approved):

PR #912, opened 2023-12-01
Automatic reconcile

achimweigel:

/area deployers
/kind enhancement
/priority 3

Add the reconcile-if-changed annotation to a root installations results in automatic reconcile operations if the spec of the installation is changed. This is required to support the use case where e.g. flux is watching installations in a git repo. Adding the reconcile annotation to these installations would result in a situation where flux is adding again this annotation when if was removed by the landscaper after every reconsiliation. 

Fixes #

**Release note**:

```other operator
- automatic reconcile if installation spec changed
```

### DIFF
--- a/pkg/landscaper/controllers/installations/controller_test.go
+++ b/pkg/landscaper/controllers/installations/controller_test.go
@@ -82,9 +82,9 @@ var _ = Describe("Installation Controller", func() {
 		})
 
 		It("should reconcile an installation with reconcile-if-changed annotation", func() {
-			// We consider an Installation in a finished state.
-			// The Installation has no reconcile annotation. Therefore, a reconciliation should have no effect.
-			// After the reconciliation, the jobID and jobIDFinished should be the same as before.
+			// We consider an Installation with a reconcile-if-changed annotation and an observed generation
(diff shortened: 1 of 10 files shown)


pkg/landscaper/controllers/installations/controller_test.go
```diff
@@ -81,6 +81,30 @@ var _ = Describe("Installation Controller", func() {
 			Expect(inst.Status.InstallationPhase).To(Equal(lsv1alpha1.InstallationPhases.Succeeded))
 		})
 
+		It("should reconcile an installation with reconcile-if-changed annotation", func() {
+			// We consider an Installation in a finished state.
+			// The Installation has no reconcile annotation. Therefore, a reconciliation should have no effect.
+			// After the reconciliation, the jobID and jobIDFinished should be the same as before.
```
robertgraeff, line 87: The comment is wrong, copied from the previous test.

robertgraeff (changes requested):

pkg/landscaper/controllers/installations/controller_test.go
achimweigel: ups - I have fixed this

robertgraeff (approved):

PR #921, opened 2023-12-08
local blob deployer

fabianburth:
This PR introduces new templating functions `getResourceKey(resourceIdentity string)` and `getResourceContent(resourceIdentity string)`. These functions work for all resources specified as part of an ocm component.

`getResourceKey()` returns a key that can be used by the helm deployer to fetch the chart corresponding to the specified resource. 
`getResourceContent()` templates the entire content of a resource into deploy execution. The format in which it is "printed" into the deploy execution, per default, is the string representation of the corresponding bytes. If the media type of the resource is `application/octet-stream`, its "printed" as a base64 encoded representation of the bytes instead. 

/area deployers
/kind task
/priority 2

Fixes #

**Release note**:

```feature user
access type independent templating function for ocm resources
```


achimweigel:
Unfortunately I could not review this PR the normal way. My browser is just not able to handle the size of it. What I have seen is that there is no documentation with examples for this. Could you perhaps add something here?

fabianburth:
> Unfortunately I could not review this PR the normal way. My browser is just not able to handle the size of it. What I have seen is that there is no documentation with examples for this. Could you perhaps add something here?

Open a new PR #958 with the corresponding documentation.

achimweigel (approved):

PR #938, opened 2024-01-02
Fix empty schema in exports

achimweigel:

Fix missing schema in export definitions. Fixes problem described [here](https://github.tools.sap/kubernetes/k8s-lifecycle-management/issues/1049). 

Fixes #

**Release note**:

```bugfix user
- fix missing schema in export definitions
```

### DIFF
--- a/pkg/landscaper/installations/exports/constructor.go
+++ b/pkg/landscaper/installations/exports/constructor.go
@@ -112,7 +112,7 @@ func (c *Constructor) Construct(ctx context.Context) ([]*dataobjects.DataObject,
 		switch def.Type {
 		case lsv1alpha1.ExportTypeData:
 			if def.Schema == nil {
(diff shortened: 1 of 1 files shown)


pkg/landscaper/installations/exports/constructor.go
```diff
@@ -111,6 +111,10 @@ func (c *Constructor) Construct(ctx context.Context) ([]*dataobjects.DataObject,
 
 		switch def.Type {
 		case lsv1alpha1.ExportTypeData:
+			if def.Schema == nil {
+				return nil, nil, fmt.Errorf("%s: schema is nil", fldPath.String())
```
reshnm, line 115:
```suggestion
				return nil, nil, fmt.Errorf("%s: schema for data export %q may not be nil", fldPath.String(), def.Name)
```

reshnm (changes requested):

reshnm (approved):

PR #956, opened 2024-01-24
Further improvements

achimweigel:

Further improvements:

- set finaliser during creation to prevent additional write operations
- remove some update operations

Fixes #

**Release note**:

```other user
- set finaliser during creation to prevent additional write operations
- remove some update operations
```


apis/core/types_execution.go
```diff
@@ -86,10 +86,6 @@ type ExecutionStatus struct {
 	// +optional
 	DeployItemCache *DeployItemCache `json:"deployItemCache,omitempty"`
 
-	// ExecutionGenerations stores which generation the execution had when it last applied a specific deployitem.
-	// So in this case, the observedGeneration refers to the executions generation.
-	ExecutionGenerations []ExecutionGeneration `json:"execGenerations,omitempty"`
```
robertgraeff, line 91: If we remove this field, do we still need the `ExecutionGeneration` structure below?

robertgraeff (changes requested):

apis/core/types_execution.go
achimweigel: I have removed this

robertgraeff (approved):

PR #985, opened 2024-02-21
added frontmatter metadata to Usage Docs

In-Ko:
Added necessary FrontMatter Metadata to all docs under "Usage"


guewa (approved):
i will push a 'make generate-docs' for you...

PR #991, opened 2024-02-28
[Draft] Add Dev Container

enrico-kaack-comp:

Adds .devcontainer including all required tools and components for a full dev container including minikube and registry.
NOTE: not yet completed.

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/go:1.22
+FROM mcr.microsoft.com/devcontainers/go:1.22-bullseye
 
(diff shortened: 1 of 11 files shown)


enrico-kaack-comp:
Performance Checks on M2 Pro 16GB:

**Local Run:**
55s Local landscaper-controller run without cluster, deleted cache
20s subsequent run without cluster, with cache

**Colima Docker DevContainer with 8CPU/12GB**
3m  Landscaper controller run with minikube
1m 7s subsequent run with miniklube

**--> 3x slower build compared to local**

Suggestion: Scripts to setup a local minikube, registry (without TLS), create a kubeconfig, create target, create context.
Then the build is as fast as possible and the local setup is easy.

PR #993, opened 2024-02-29
Add OCM Verify to installation

enrico-kaack-comp:

Implements OCM verify for signature verification. 
Usage details can be found in docs/usage/SignatureVerification.md.

Fixes #

**Release note**:

```feature user|operator
- adds signature verification capabilities. IMPORTANT: In the current state, the landscaper can only verify the signature and content when the installation is created/modified. 
Later controllers such as deployer will access the resources WITHOUT further verification. This implies that a potential malicious manipulation between creating the installation (and verifying the signature) and subsequent resource access can NOT be detected/prevented. 
Therefore, it is advisable to use this feature with a trusted repository only.
```


apis/core/types_installation.go
```diff
@@ -47,6 +47,8 @@ type InstallationSpec struct {
 	// +optional
 	Context string `json:"context,omitempty"`
 
+	Verification *Verification `json:"verification,omitempty"`
```
reshnm, line 50: I would rather not have that specified in the Installation directly but have it in the Landscaper Context.
Pros:
- It can be used by mutliple installations
- It only has to be specified once
- The PKI is bound to the OCI registry in which the component is retrieved from, which is also specified in the Context.

reshnm (changes requested):

apis/core/types_installation.go
enrico-kaack-comp: Good point, i will add this. Thank you

docs/usage/SignatureVerification.md
```diff
@@ -0,0 +1,73 @@
+# Signature Verification
+
+Landscaper can verify the signature of a component version and only continue with the installation if the signature is valid and matches the content.
+
+> **IMPORTANT**: In the current state, the landscaper can only verify the signature and content when the installation is created/modified. Later controllers such as deployer will access the resources WITHOUT further verification. This implies that a potential malicious manipulation between creating the installation (and verifying the signature) and subsequent resource access can NOT be detected/prevented. Therefor, it is advisable to use this feature with a trusted repository only.
```
achimweigel, line 5:
```suggestion
> **IMPORTANT**: In the current state, the landscaper can only verify the signature and content when the installation is created/modified. Later controllers such as deployer will access the resources WITHOUT further verification. This implies that a potential malicious manipulation between creating the installation (and verifying the signature) and subsequent resource access can NOT be detected/prevented. Therefore, it is advisable to use this feature with a trusted repository only.
```

docs/usage/SignatureVerification.md
```diff
@@ -0,0 +1,73 @@
...
+### Landscaper Config Signature Verification Enforcement Policy
+
+In the landscaper config, a verification enforcement policy can be specified:
+```yaml
+useOCMLib: true
+signatureVerificationEnforcementPolicy: DoNotEnforce # DoNotEnforce(DEFAULT)|Enforce|Disabled
+```
+The following values are possible:
+- **DoNotEnforce** (DEFAULT): does not enforce a landscaper-wide policy. Each installation can specify a `spec.verification` key to enable the verification for it.
+- **Enforce**: enforces all installations to be verified and therefore all installation need the `spec.verification` key with provided verification options.
```
achimweigel, line 25:
```suggestion
- **Enforce**: enforces all installations to be verified and therefore all installations need the `spec.verification` key with provided verification options.
```

docs/usage/SignatureVerification.md
```diff
@@ -0,0 +1,73 @@
...
+- **Disabled**: explicitly disable all verifications for all installations even if a installations provides verification information with `spec.verification` key. **It is advised to use this option for debugging only.**
```
achimweigel, line 26:
```suggestion
- **Disabled**: explicitly disables all verifications for all installations even if an installation provides verification information with `spec.verification` key. **It is advised to use this option for debugging only.**
```

docs/usage/SignatureVerification.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Configure Verification
+
+Necessary verification information is provided in the installation and the context.
+In the installation, `spec.verification.signatureName` needs to be set to the name of the signature in the component version:
```
achimweigel, line 31: Isnt't it the name of the signature in the context?

docs/usage/SignatureVerification.md
```diff
@@ -0,0 +1,73 @@
...
+# OR
+    publicKeySecretReference:
+      name: secret-publickey
+      namespace: default
+      key: key
+```
+The referenced secret contains the PEM encoded public key or ca certificate.
+
+Which option to use depends on how the component version was signed. The `publicKeySecretReference` maps to the `--public-key` of the ocm cli command `ocm verify` and the `caCertificateSecretReference` maps to the `ca-cert` option.
+If the signature field of the component version also contains a certificate of the signer, the `caCertificateSecretReference` with the issuer certificate of the signer.
```
achimweigel, line 73: Would it perhaps be helpful, if we show both cases with small examples?

pkg/landscaper/controllers/installations/controller.go
```diff
@@ -367,6 +368,23 @@ func (c *Controller) initPrerequisites(ctx context.Context, inst *lsv1alpha1.Ins
 		return nil, lserrors.NewWrappedError(err, currOp, "SetupRegistries", err.Error())
 	}
 
+	if verify.IsVerifyEnabled(inst, c.LsConfig) {
```
achimweigel, line 371: If you check the call hierarchy of the method initPrerequisites, this method is called twice during the processing of an installation, in the methods handlePhaseInit and handlePhaseCompleting. Would it perhaps be a possibility to verify the component just in the first call, e.g. by calling the methods with some boolean parameter to determine if verification should be executed or not.

achimweigel (changes requested):

docs/usage/SignatureVerification.md
enrico-kaack-comp: yes, in the context we have a map of signatureName: {} that allows multiple signature names to be defined in the context. For the installation, the user must pick which signatureName should be chosen.

achimweigel: But you wrote "needs to be set to the name of the signature in the component version" instead of "needs to be set to the name of the signature in the context" :-)

docs/usage/SignatureVerification.md
enrico-kaack-comp: I suggest this to be documented in more detail in the ocm lib, since we only pass the paramter to the ocm lib.

pkg/landscaper/controllers/installations/controller.go
enrico-kaack-comp: I added the suggested change. I think the method initPrerequisites should be refactored instead of adding a flag though.

docs/usage/SignatureVerification.md
enrico-kaack-comp: Now I understand, the name of the signature in the component version == name of the signature in the context == name of the signature in the installation.

docs/usage/SignatureVerification.md
achimweigel: I would at least give an example secret and a link how to create a PEM encoded public key or ca certificate. It should be easy for everyone in our team to setup a corresponding example without all the exploration you did for this.

pkg/landscaper/controllers/installations/controller.go
achimweigel: Perhaps I'm totally confused but you call the method always with `runVerify==false` and therefore a verification is never executed.

enrico-kaack-comp: could you please check again, I call it with true the first time
https://github.com/gardener/landscaper/pull/993/files#diff-766e33e443b4adbcff54a27f95dca2e50e0be398f347541bae6b62896dfd716bR239

achimweigel: You are totally right. I missed that the method was called twice in the first method, one time with true.

achimweigel (approved):

reshnm (approved):

PR #1003, opened 2024-03-06
Update guided tour

achimweigel:

Update guided tour.

Fixes #

**Release note**:

```doc user
- Update guided tour
```


docs/guided-tour/README.md
```diff
@@ -25,20 +27,12 @@ sidebar_position: 1
...
-
 ## How to follow the Tour
 
 The Guided Tour consists of several chapters, some of which build on each other. In order to get the most out of it, 
-you should be following the Guided Tour in this sequence. Another important point is, that this documentation
-is not optimal with respect to getting an overview about the used example resources in the Guided Tour. Therefore,
-it might be better to read it in the [original github repository](https://github.com/gardener/landscaper/tree/master/docs/guided-tour).
+you should be following the Guided Tour in this sequence. The Guided Tour is not presented very well
+if preprocessed with some documentation tooling. It might be better to read it with your browser in the 
+[original github repository](https://github.com/gardener/landscaper/tree/master/docs/guided-tour).
```
reshnm, line 35: I don't really understand this. Who is preprocessing with some documentation tooling? How is this impacting the Landscaper user?

docs/guided-tour/blueprints/echo-server/README.md
```diff
@@ -74,10 +74,10 @@ component. In detail, the connection is the following:
...
   ```
   
-> **_NOTE:_** Since Kubernetes does not support OCM (yet ;) ), we need the *oci reference* of the container image, here.
-> Consequently, to actually use this component with landscaper, the container image that has to be deployed in a pod 
-> cannot be embedded into the component as a local blob (though it make sense to do so, as an intermediate step during 
-> transport of the component).
+> **_NOTE:_** Since Kubernetes does not support OCM, we need the *oci reference* of the container image, here.
+> Consequently, to actually use this component version with landscaper, the container image that has to be deployed in 
+> a pod cannot be embedded into the component as a local blob (though it make sense to do so, as an intermediate step 
+> during transport of the component).
```
reshnm, line 80: Does a reader really know what a local blob is in the context of ocm? I don't think so.

I think we need to give the user a more detailed description of how they have to specify OCI image resources in the component version so that they will be consumable by k8s. Could also be part of the ocm.software docs and be referenced here.

docs/guided-tour/blueprints/external-blueprint/README.md
```diff
@@ -27,30 +27,15 @@ We have uploaded the blueprint
...
-component as a so-called local blob. Then, the *blob* directory in the file system representation would not have been 
-empty. It would have contained the blueprint (typically as a tar archive), as demonstrated 
-[here](./components/local-blob-component).
-
-The most convenient way to generate such components is through corresponding *component configuration files*. The
-*component configuration file* used to create the component for this example is shown 
-[here](./config-files/components.yaml): 
+The most convenient way to generate such component versions is through corresponding *component configuration files*. 
+The *component configuration file* used to create the component version for this example is shown
+[here](./config-files/components.yaml):
```
reshnm, line 39: I think we should cross reference https://ocm.software/docs/guides/getting-started-with-ocm

docs/guided-tour/blueprints/external-blueprint/README.md
```diff
@@ -67,16 +52,46 @@ components:
...
-component to an OCI registry can be found [here](./commands/component.sh).
+The commands used to create the actual component version based on the *component configuration file* and to upload this
+component version to an OCI registry can be found [here](./commands/component.sh). The first command creates
+a file system representation of the component version and the second uploads this to the registry:
 
->**Tip:** If you need the same resource in multiple components, instead of copy-pasting, you might want to use 
-> [*resource configuration files*](https://ocm.software/docs/guides/getting-started-with-ocm/#using-a-resources-file).
+```
+ocm add components --create --file ../tour-ctf ../config-files/components.yaml
+ocm transfer ctf --enforce ../tour-ctf eu.gcr.io/gardener-project/landscaper/examples
```
reshnm, line 61: Why the enforce flag?

docs/guided-tour/blueprints/external-blueprint/README.md
```diff
@@ -67,16 +52,46 @@ components:
...
+and uploading of the component version is the same as before. If you download this component version as before
+you get [this](components/local-blob-component) file system representation. Now the component descriptor references
+the blueprint as a local blob. This local blob is contained in the file system representation of the component
+version in the [blobs folder](components/local-blob-component/blobs).
+
+To embed a blueprint into a component version as a local blob is usually the preferred approach.
+
+For more details about component versions check out the [OCM Getting Started Guide](https://ocm.software/docs/guides/getting-started-with-ocm/). In the OCM documentation
+you also find more about the advantages of using components to describe your artefacts like transportability, signing, 
+scanning etc.
```
reshnm, line 90: Why are we advertising the concept of having external blueprints? Yes, this is possible, but nowhere used in a real scenario.
I would prefer that we just describe the "standard" way of embedding a blueprint ans a local blob.

I would also like to have examples embedded instead of links. The reader has to jump back and forth.
Why not give an inline example how a bloueprint has to be specified in a component version?

docs/guided-tour/blueprints/external-blueprint/README.md
```diff
@@ -93,16 +108,15 @@ blueprint:
 
 - The field `context` contains the name of a custom resource of kind [Context](../../../usage/Context.md) in the same
   namespace as the Installation on the Landscaper resource cluster. [Our Context resource](./installation/context.yaml)
-  contains the information in which registry the component descriptor and blueprint are stored.
+  contains the information in which registry the component version with the component descriptor is stored.
```
reshnm, line 111: Here it would also nice to have an inline example of a context and how the repository context is specified.

docs/guided-tour/blueprints/external-blueprint/README.md
```diff
@@ -93,16 +108,15 @@ blueprint:
...
   component descriptor in that registry.
 
-- The [component descriptor](./component-archive/v2-external/component-descriptor.yaml) contains a list of resources,
+- The [component descriptor](components/component/component-descriptor.yaml) contains a list of resources,
   each of which has a name. Field `blueprint.ref.resourceName` in the Installation specifies the name of the blueprint
-  resource in the component descriptor. Thereby, it is completely transparent for the installation whether the component
-  references the blueprint as an external resource or embeds it as a local blob. 
-
+  resource in the component descriptor. Thereby, it is completely transparent for the installation whether the 
+  component version references the blueprint as an external resource or embeds it as a local blob.
```
reshnm, line 119: Same as above. Confuses the reader.

docs/guided-tour/blueprints/helm-chart-resource/README.md
```diff
@@ -3,39 +3,44 @@ title: Helm Chart Component
 sidebar_position: 2
 ---
 
-# Helm Chart Resources in the Component
+# Helm Chart Resources in the Component Version
 
-Let's look at an example in which a Helm Chart is referenced from a Component.
+Let's look at an example in which a Helm Chart is referenced from a component version.
```
reshnm, line 8:
```suggestion
Let's look at an example in which a Helm Chart is referenced out of a component version.
```

docs/guided-tour/blueprints/helm-chart-resource/README.md
```diff
@@ -3,39 +3,44 @@ title: Helm Chart Component
...
 ```
 
-Using this oci reference, the landscaper is able to fetch the helm chart and deploy it correspondingly.
+Using this oci reference, the landscaper is able to fetch the helm chart and deploy it correspondingly. This approach 
+has the disadvantage, that the location of the chart is hidden in the blueprint which is an unknown resource type
+of OCM. Therefore, if OCM component versions are used for transport, scanning etc. such hidden artefact could not be 
+identified as part of the component version and are just skipped. A better approach is to specify such artefacts in 
+the component version itself such that they are known on the component level and just reference the artefacts in 
+the blueprint via their entries in the component descriptor of the component version. We will show this in the 
+following with the helm chart artefact.
```
reshnm, line 27: I would remove this part altogether. Just document one approach.
To my understanding this way of referencing charts is deprecated and should not be used.

docs/guided-tour/blueprints/helm-chart-resource/README.md
```diff
@@ -85,8 +90,7 @@ components:
           helmRepository: https://example.helm.repo.com/landscaper
 ```
 
-
-Again, if you were to prefer to embed the blueprint and the helm chart in the component as a local blob, instead of an 
+Again, if you prefer to embed the blueprint and the helm chart in the component as a local blob, instead of an
```
reshnm, line 93: This is not correct. A resource being a local blob is not determined if it is specified by `access` or `input`.
You can use `input` and not have a local blob.

reshnm (changes requested):
In general I would prefer to have more inline embedded code that doesn't interrupt the reading flow. Of course you can have the links to the complete files at the same time.

I think we should just describe best practices. For me it makes no sense to handle external blueprints. This hasn't been used in any real use case. Usally blueprints are shared via component references.

robertgraeff:
> In general I would prefer to have more inline embedded code that doesn't interrupt the reading flow. Of course you can have the links to the complete files at the same time.
> 
> I think we should just describe best practices. For me it makes no sense to handle external blueprints. This hasn't been used in any real use case. Usally blueprints are shared via component references.

We put this on the todo list https://github.tools.sap/kubernetes/k8s-lifecycle-management/issues/1172

docs/guided-tour/README.md
robertgraeff: We have removed this sentence.

docs/guided-tour/blueprints/echo-server/README.md
robertgraeff: We will rework the three blueprints example and combine them into one example. It is on our follow-up list.

docs/guided-tour/blueprints/external-blueprint/README.md
robertgraeff: It is on our follow-up list.

docs/guided-tour/blueprints/external-blueprint/README.md
robertgraeff: To overwrite an already existing version.

docs/guided-tour/blueprints/external-blueprint/README.md
robertgraeff: It is on our list to replace all external blueprints by local ones.

docs/guided-tour/blueprints/external-blueprint/README.md
robertgraeff: Is on our list.

docs/guided-tour/blueprints/helm-chart-resource/README.md
robertgraeff: Fixed

docs/guided-tour/blueprints/helm-chart-resource/README.md
robertgraeff: We explain this during the rework of the chapter. We should probably link to https://ocm.software/docs/guides/input_and_access/

docs/guided-tour/blueprints/helm-chart-resource/README.md
robertgraeff: We replace the three example in the blueprints section by one example. In this example all resources (blueprint, chart, image) will be in the component descriptor, the blueprint as local resource. As the other points this is on the follow-up list https://github.tools.sap/kubernetes/k8s-lifecycle-management/issues/1172.

robertgraeff (approved):

PR #1008, opened 2024-03-12
integration test refactoring

reshnm:

This pr removes severeal layers of bash and python script calls until the integration test is finally started.
This hopefully reduces the complexity of how the integration tests are executed.
This PR fouces on how the integration test is started and executed in the context of a pipeline job.
I tried the best to not break some local test execution scenarios, but they might be broken. We should fix the along the way when they are encountered.

The second improvement of this PR is that the integration test is now executed in a way that is very close to our real LAAS setup. One target shoot cluster running the Landscaper controllers, while all Landscaper resources are placed on a seperate resource shoot cluster.

When installing the Landscaper I have tried to use the ocm tooling as much as possible so that we can make sure that the component version is working as a single source of truth.

Fixes #

**Release note**:

```feature operator
Run Landscaper integration test on a seperate target shoot cluster and resource shoot cluster.
```


achimweigel (approved):
I trust you :-)

PR #1021, opened 2024-03-19
Inform about panics

achimweigel:

Inform about panics

Fixes #

**Release note**:

```feature operator
Inform about panics
```


pkg/utils/critical_problems_handler.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return nil
+}
+
+func (r *criticalProblemsHandler) ProvokePanicIfAnnotated(object *metav1.ObjectMeta) {
+	if kubernetes.HasAnnotationWithValue(object, v1alpha1.LandscaperDomain+"/throwpanic", "active") {
+		a := 1
+		b := 2
+		b = a + b
+		c := a / (4 - b - a)
```
Diaphteiros, line 100: Is this meant to cause a panic by dividing by zero without the linter noticing you trying to divide by zero?
This needs at least a comment with an explanation, but I wonder what prevents us from just calling `panic(...)` here instead of doing this?

pkg/utils/panic_handler.go
```diff
@@ -33,21 +35,33 @@ func HandlePanics(ctx context.Context, result *reconcile.Result) {
 
 			if err.Error() == "runtime error: invalid memory address or nil pointer dereference" {
 				logger.Error(err, "Recovered from a nil pointer dereference or invalid memory address error")
```
Diaphteiros, line 37: Would it be possible to add this error message (or even better, the original panic error) to the CR? I think it would be helpful to not only see 'pod X crashed 5 minutes ago', but also get some information about why it crashed. Of course, it can be read from the logs, but if I check the CR anyway, it would be easier to get the information from there.

Diaphteiros (changes requested):

pkg/utils/critical_problems_handler.go
achimweigel: I have removed the complete method. This was only a leftover for some tests.

pkg/utils/panic_handler.go
achimweigel: This is a good idea. I have added a description containing the error text. If required, we can later further extend the entries.

Diaphteiros (approved):

PR #1046, opened 2024-03-27
fix context crd (temporarily disabled openapi spec generation)

Diaphteiros:
**Release note**:

```bugfix user
Fixed a bug that resulted in a wrongly generated CRD for the `Context` resource.
```


hack/generate-code.sh
```diff
@@ -50,33 +50,33 @@ kube::codegen::gen_helpers \
...
+#   --extra-pkgs "$API_MODULE_PATH/deployer/mock/v1alpha1" \
+#   --extra-pkgs "github.com/gardener/component-spec/bindings-go/apis/v2" \
+#   --extra-pkgs "k8s.io/api/core/v1" \
+#   --extra-pkgs "k8s.io/apimachinery/pkg/apis/meta/v1" \
+#   --extra-pkgs "k8s.io/apimachinery/pkg/api/resource" \
+#   --extra-pkgs "k8s.io/apimachinery/pkg/types" \
+#   --extra-pkgs "k8s.io/apimachinery/pkg/runtime" \
+#   --report-filename "$src_path/$API_MODULE_PATH/openapi/api_violations.report" \
+#   --update-report \
+#   --boilerplate "${PROJECT_ROOT}/hack/boilerplate.go.txt"
```
reshnm, line 80: Is this commented section intentional?

reshnm (changes requested):

reshnm:
It seems that we had no test to cover this use case. Wouldn't it make sense to have at least one test that uses this field so that we know earlier in the future?

hack/generate-code.sh
Diaphteiros: Kind of. For some reason that I have not figured out yet, the command breaks. And since the bug fixed by this PR is currently affecting one of our customers, we might want to merge this PR before I have figured out how to get the openapi generation going again.

Diaphteiros:
> It seems that we had no test to cover this use case. Wouldn't it make sense to have at least one test that uses this field so that we know earlier in the future?

In theory, I agree. In practise, I doubt that we will mess with the CRD generation again anytime soon, so it is probably not that important. OK, the missing test is causing problems now, but we had the previous CRD generation in place for several years, so unless the code-generator support is dropped in the near future, we will not touch this for years again, I think.

hack/generate-code.sh
reshnm: 👍

reshnm:
> > It seems that we had no test to cover this use case. Wouldn't it make sense to have at least one test that uses this field so that we know earlier in the future?
> 
> In theory, I agree. In practise, I doubt that we will mess with the CRD generation again anytime soon, so it is probably not that important. OK, the missing test is causing problems now, but we had the previous CRD generation in place for several years, so unless the code-generator support is dropped in the near future, we will not touch this for years again, I think.

But how do we know this fix really works? Have you done some local testing?
I would feel much more confident by having a least a small simple test before shipping the fix.

reshnm (approved):

Diaphteiros:
I tested it locally. Let's ship the fix and then create a test and fix the openapi generation afterwards.

PR #1055, opened 2024-04-02
ocm config support

fabianburth:
This PR enables the landscaper to consume ocm config data that is usually configured in ocm config files (located at $HOME/.ocmconfig per default, if you have ocm-cli installed). For the landscaper, the ocm config data has to be specified in a config map which can be referenced in the landscaper context object.
The primary purpose for enabling this is that it allows the user to specify multiple ocm repositories (called resolvers in the config file) from which the component specified in the installation and the components referenced by that component may be resolved (thus, referenced components do not necessarily have to be in the same repository, as long as the other repository is also specified as resolver).

This implementation also allows to configure credentials within this configuration data, although it is not encouraged to specify credentials in clear text in the config map. But one may specify a reference to the credentials stored somewhere else. A currently already supported method is to reference credentials stored in vault.

Fixes #

**Release note**:

```feature user

```


fabianburth:
added documentation in the usage documentation of the landscaper context

robertgraeff (approved):

PR #1083, opened 2024-04-18
Update to go 1.22

guewa:

Fixes #

**Release note**:

```other dependency
Update to go 1.22
```


achimweigel (changes requested):
I just run into the problem, that make generate is not working with golang version 1.22. Johannes told that this is a known problem. We should check this before we merge this PR.

achimweigel (approved):
Make generate is also working locally. I guess we can merge the PR.

PR #1090, opened 2024-04-23
refactor CRD embedding

Diaphteiros:
The generated CRDs are moved from `pkg/landscaper/crdmanager/crdresources` to `apis/crds/manifests` and added a `CRDs` function to the embedding package (`apis/crds`).

The main advantage of this change is that it becomes significantly easier to access the CRDs from other projects.

Based on #1089, which should be merged before this one.

**Release note**:

```breaking developer
The CRD manifests have been moved from `pkg/landscaper/crdmanager/crdresources` to `apis/crds/manifests`. The new `apis/crds` package contains a `CRDs` function which provides easy access to the CRDs to other go projects.
```


achimweigel (changes requested):
Could you perhaps run the integration tests?

achimweigel (approved):

PR #1107, opened 2024-05-03
Extend cache to component versions and blueprints

achimweigel:

Extend cache to component versions and blueprints.

Fixes #

**Release note**:

```feature user
Extend cache to component versions and blueprints
```


pkg/utils/resource_cache/resource_cache.go
```diff
@@ -0,0 +1,247 @@
...
+	resourceBytesCompressed []byte
+	timestamp               time.Time
+}
+
+func (c *CacheEntry) GetEntries() ([]byte, time.Time) {
+	return c.resourceBytesCompressed, c.timestamp
+}
+
+var resourceCache *ResourceCache
+var getInstanceLock sync.RWMutex
```
fabianburth, line 44: I think this could be simplified using a once object.

fabianburth:
Please let me check what could be covered with the ocm lib caching before moving forward with this implementation.

achimweigel:
> Please let me check what could be covered with the ocm lib caching before moving forward with this implementation.

@fabianburth it would be perfect if this could be handled by the OCM lib. Do you already know when you have time to check this. Steampunk is waiting on it. The OCM caches must not be a blackbox for the Landscaper. Therefore the following requirements are important:

- It must be possible, on an Installation basis, to control if the caches should be used or not. 
- It must be possible to define the maximal size in bytes of the caches
- It must also be obvious how the caches work, e.g. how does the removal strategy looks like, what is cached with respect to the operations used by the Landscaper etc.

PR #1193, opened 2024-07-30
Fix helm log function

robertgraeff:

This PR fixes how the helm log function fills a list of values into a format string. The first placeholder of the format string was filled with the complete value list, instead of only the first value. All other placeholder did not get a value. As a result, the log messages looked for example like this:

```
"beginning wait for [16 598623897521] resources with timeout of %!v(MISSING)"
```

Fixes #

**Release note**:

```bugfix user
- Fix helm log function
```


achimweigel:
@robertgraeff are you sure that this is required? If you look at my change before, the call was exactly the same.

achimweigel (approved):

PR #1257, opened 2024-11-04
Fix Unmarshaling of Target Config

robertgraeff:

Fixes #

**Release note**:

```bugfix user
- Bugfix concerning Targets that reference a Secret by name and key
```

### DIFF
--- a/pkg/utils/clusters/shoot_client.go
+++ b/pkg/utils/clusters/shoot_client.go
@@ -6,7 +6,6 @@ package clusters
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/utils/clusters/shoot_client.go
```diff
@@ -56,7 +56,7 @@ func NewShootClientFromTarget(ctx context.Context, gardenTarget *v1alpha1.Target
 	}
 
 	targetConfig := &targettypes.KubernetesClusterTargetConfig{}
-	err = json.Unmarshal([]byte(resolvedTarget.Content), targetConfig)
+	err = yaml.Unmarshal([]byte(resolvedTarget.Content), targetConfig)
```
reshnm, line 59: I guess this is fine, since the yaml parser should also be able to parse json (?).
However, I believe we need to also fix the secret resolving in `controller-utils/pkg/landscaper/references.go`. The function `ResolveSecretReference` should return the same format for when a key is specivied vs. there is no key specified. What do you think?

reshnm (changes requested):

reshnm (approved):

